PR #145, opened 2014-09-20
Implementing the 'DynamicShould.HaveProperty' functionality.

chaitanyagurrapu:


JakeGinnivan:
@chaitanyagurrapu am just revisiting stuff in preparation for coreclr and this code doesn't actually support if there is no stackframe. This happens if you compile in release mode, using this assertion with release mode compilation will cause shouldly to throw. Do you think you could have a look into it?

chaitanyagurrapu:
Sure, I will have a look. I will have to first get up to speed on the new core clr / dnr architectures coming up. When do you need it by @JakeGinnivan ?

JakeGinnivan:
@chaitanyagurrapu things are pretty much ready to go for the next release, the `DynamicShould` tests are all being skipped at the moment. Do you think you will get a chance to look into it?

chaitanyagurrapu:
Hi @JakeGinnivan. I've been having a look at this again. The code is very different from the last time i've had a look at it.

I've submitted a PR with an initial attempt that should fix the unit tests in debug mode. Have a look and see what you think.

In debug mode, when I try to run all the tests in the solution by right clicking the solution and choosing run all unit tests. It seems the unit tests are hanging. See screenshot below.

image

I cannot compile the solution in Release mode. The Shouldly.Portable project doesn't build. This is of course what you were mentioning above. But the stack trace isn't a specific dependency just for DynamicShould. It is used for all the shouldly methods. It is the one that gets us the file name and line number. I will see what alternatives we have, but the side effect might be far reaching. In the mean time if you have any idea why my unit tests are hanging in debug mode, it will very helpful. :-)

Issue #41, opened 2012-04-02
Display detailed diff for string comparisons when calling ShouldBe


xerxesb:
This task is to enhance the message reporting of ShouldBe(string) method to show more information about the string and where it differed from the target.

The basic idea (from my recollection) was taken from Phil Haack's blogpost on comparing strings un unit tests. (http://haacked.com/archive/2012/01/14/comparing-strings-in-unit-tests.aspx)

He has a reference implementation which (with some love) should slip into Shouldly. (https://gist.github.com/1610603)

This task will possibly require doing the refactor to the ShouldlyMessage/ShouldlyMessageGenerator i've been putting off for some time (ie: All messages should be routed through an appropriate generator and the ShouldlyMessage shouldn't do any message output).

JakeGinnivan:
Done, thanks to @chaitanyagurrapu

PR #163, opened 2014-12-20
Refactoring the difference highlighting 

chaitanyagurrapu:
Refactored the difference highlighting system to be more centered around the 'TestEnvironment' which is our context. This makes it similar to the pattern that is used for the message generators and allows us to add new difference highlighters easily. This is an interim step towards adding issue #41 . With the 'TestEnvironment' being passed in rather than just the 'Actual' and 'Expected' values, it should be possible to tell whether or not to ignore case sensitivity for example. Basically, having the whole `TestEnvironment` class should allow us to write richer highlighters .

As an aside @JakeGinnivan , we should consider renaming the `TestEnvironment` class to be `AssertionContext`, since that is what it really does - package up all the nitty gritty details of the shouldly assertion.


JakeGinnivan:
I like the change in general.

Don't think we need the ITestEnvironment and mocking framework. We can simply do `new TestEnvironment { Actual = , Expected = }`

chaitanyagurrapu:
I did try avoiding it, but couldn't find an easy way. The problem is the tests in the "DifferenceHighlighterHelpersTests" class. New'ing up a TestEnvironment doesn't work because in the constructor it tries to walk up the stack trace to find the location of the test failure. But the test hasn't really failed yet. Have a look at this test ...

```c#
[Test]
public void HighlightDifferencesBetween_IntegerListsSameExceptOne_HighlightsDifference()
{
    var mockTestEnvironment = MockHelper.GetMockTestEnvironment(new[] { 2, 2, 3 }, new[] { 1, 2, 3 });
    DifferenceHighlighterExtensions.HighlightDifferencesBetween(mockTestEnvironment.Object).ShouldBe("[*2*, 2, 3]");
}
```

If i replaced it with  ... 

```c#
[Test]
public void HighlightDifferencesBetween_IntegerListsSameExceptOne_HighlightsDifference()
{
    var mockTestEnvironment = new TestEnvironment(new[] { 2, 2, 3 }, new[] { 1, 2, 3 });
    DifferenceHighlighterExtensions.HighlightDifferencesBetween(mockTestEnvironment.Object).ShouldBe("[*2*, 2, 3]");
}
```

On the first line, it already starts walking up the stack trace looking for the ShouldlyMethod that caused the test to fail. But the test hasn't failed yet. Basically this loop in the TestEnvironment constructor...

```c#
...
while (shouldlyFrame == null || IsShouldlyMethod(currentFrame.GetMethod()))
{
    if (IsShouldlyMethod(currentFrame.GetMethod()))
        shouldlyFrame = currentFrame;
...
```

at some stage bottoms out, because TestEnvironment only expects to be constructed when one of the Shouldly methods "chucks a wobbly".

I tried modifying the constructor to be able to handle this case, but it was starting to become ugly. Which of course is the exact problem mocking can help us solve.

chaitanyagurrapu:
Sure. In that case I will leave the interface in and create a test context. Will update the PR soon ...

chaitanyagurrapu:
Done @JakeGinnivan . Very interesting blog post btw. I found Moq syntax to be a bit rough around the edges as well. So, inspired by Shouldly, i've started a project to make Moq more user friendly (https://github.com/chaitanyagurrapu/ArrangeMock) :-)

JakeGinnivan:
I rebased and squashed the two commits together which means moq never ends up in the repo history and merged via b5584c0c03e67c5862f225be74b6e2d518f32e19

Nice change

JakeGinnivan:
ArrangeMock is interesting. I find Fluent interfaces quite verbose and personally not a huge fan, but know plenty of people who love them.

It will take *a lot* of convincing for me to switch from nSubstitute :P

chaitanyagurrapu:
Fair enough. ArrangeMock only tries to make Moq's existing fluent interface a bit more nicer, not adding a whole new one.

Having said that, IMHO fluent interfaces are nice when there is a "grammer" that needs to be enforced.  i.e, you have to call method `a` and only then you can call `b,c or d` etc. As a consumer I find it quite helpful to type a dot and at each level see the next set of operations that are available to me. It's nice to be able to follow that yellow brick road through an API rather than having to grok it in its entirety. But I suppose beauty is in the eye of the beholder. :-)

PR #165, opened 2015-01-09
Show count when .ShouldBeEmpty() fails for IEnumerable

cottsak:
I'm finding the count is what I really want to know, rather than the contents of the array. The contents of said array are not useful anyway when it's not a primitive type (where .ToString() shows a useful value).

This will show the number of items before the array content.

Perhaps this should be elsewhere too? Thoughts?

### DIFF
--- a/src/Shouldly.Tests/ShouldBeEmpty/ArrayScenario.cs
+++ b/src/Shouldly.Tests/ShouldBeEmpty/ArrayScenario.cs
@@ -11,7 +11,7 @@ namespace Shouldly.Tests.ShouldBeEmpty
 
         protected override string ChuckedAWobblyErrorMessage
         {
-            get { return "new[] { 1 } should be empty but was [1]"; }
+            get { return "new[] { 1 } should be empty but had 1 item(s) and was [1]"; }
         }
 
         protected override void ShouldPass()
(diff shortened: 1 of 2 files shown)


JakeGinnivan:
Makes sense to me. Lets pull the count out as a variable and if it is 1 drop the (s) :)

cottsak:
Kool bro.

Why extract the count to a var? Isn't that over engineering? Does the `(s)` bother you that much?

JakeGinnivan:
Reads nicer, from memory most of shouldly does that sort of thing.

cottsak:
I can be a team player. ;)

Good to go now.

chaitanyagurrapu:
Do you mind updating the documentation as well please (ReadMe.md)? Otherwise, it will start to become outdated. :-)

JakeGinnivan:
Good call @chaitanyagurrapu 

@cottsak you able to do that (just the example in the readme with the new message). Just saves me doing it :P

JakeGinnivan:
dw, I will do it. Already got you to update this once

JakeGinnivan:
Thanks matt!

cottsak:
@JakeGinnivan I would have done it.. just don't work on weekends! :p

PR #178, opened 2015-02-22
Issues/174   improve should contain error message

urig:
Implemented a new MessageGenerator so any calls to ShouldContain() where the expected type has the default impl. of ToString get a message ending with "...but did not". If the expected type has overridden ToString(), then the actual enumerable is printed using ToString() as before.


JakeGinnivan:
Looks good, I was just updating the docs with this change and thinking that we should probably do the Predicate overload and also the `ShouldNotContain` overloads as well.

Did you want to do that in the same Pull request?

urig:
Thanks Jake :). What do you mean by predicate overload? Same PR is cool by me - how should I let you know when ready?

JakeGinnivan:
There is an overload where you can go `.ShouldContain(i => i.Id == 1)` our something. Signature should be `.ShouldContain<T>(Func<T, bool> predicate)`. There should be matching overloads for ShouldNotContain.

Just post a comment when ready to let me know when done.

Sent from my Windows Phone
________________________________

urig:
I've merged master into my branch and now it doesn't build: 

>'object' does not contain a definition for 'Inspect' and no extension method 'Inspect' accepting a first argument of type 'object' could be found (are you missing a using directive or an assembly reference?)	D:\dev\shouldly\src\Shouldly\MessageGenerators\ShouldContainATypeWithDefaultToStringMessageGenerator.cs	37	83	Shouldly

Am I doing something wrong here?

urig:
@JakeGinnivan now it won't compile mostly on account of missing 'ApprovalTests' and 'Mono'?

urig:
@JakeGinnivan ready for review.
A question if you will - In my message generator I'm testing:
```
TypeHasDefaultToString(context.Expected.GetType());
```
For the predicate flow, this works "by accident" as Expected is not the same type as <T>, rather it is the expression of the predicate. Thus false is returned and the flow ends up in ShouldlyMessage.CreateExpectedErrorMessage().

Do you know if I can instead get a <T> directly through reflection (starting from the context)? That way I can genuinely tell if <T> has a default ToString() impl both for regular and for predicate forms - I think this would make for a more obvious and therefor readable implementation.

JakeGinnivan:
Sorry about that, I merged a refactoring pull request which would just caused your issues.

Sent from my Windows Phone
________________________________

JakeGinnivan:
I will try look at this tomorrow morning, need to look at it properly to answer your question.

Sent from my Windows Phone
________________________________

urig:
Nag :)

JakeGinnivan:
Am currently away at Code on the Sea. The WiFi on the ship is 75c a minute so likely I will get to this next week once I have internet which is not extortion or WiFi at a bar in Nassau :)

Sent from my Windows Phone
________________________________

Issue #150, opened 2014-11-11
ShouldAllBe message generator


JakeGinnivan:
Currently error message is `new[]{1,2,3} should all be an element satisfying the condition ((x + 4) < 7) but does not`

It should be `new[]{1,2,3,4} should satisfying the condition ((x + 4) < 7) but [3] does not` or `new[]{1,2,3} should satisfying the condition ((x + 4) < 7) but [3,4] does not`

cottsak:
`new[]{1,2,3,4} should satisfying` or `new[]{1,2,3,4} should satisfy`?

JakeGinnivan:
`satisfy` is better

cottsak:
@JakeGinnivan I'll try and jump in on this one (GitHub membership is all mine!!)

cottsak:
Fixed in https://github.com/shouldly/shouldly/pull/181

JakeGinnivan:
Will be released in next (>2.4.0) release

PR #181, opened 2015-03-11
fix ShouldAllBe message generator

cottsak:


JakeGinnivan:
Looks good

JakeGinnivan:
Should it be `do not` rather than `does not`

JakeGinnivan:
`return "new[] { 1, 2, 3 } should satisfy the condition (x < 2) but [2,3] does not"`

Does implies something you passed, do implies those are the items which do not match?

cottsak:
@JakeGinnivan yep! I like "do" better

JakeGinnivan:
Great

dennisroche:
Fixes https://github.com/shouldly/shouldly/issues/150

PR #188, opened 2015-03-16
Introduced client-friendly exception name (client being the folks that p...

### LINKED ISSUE #179: Client-friendly exception names
On a couple of clients now we've had to explain what a Wobbly is and how it getting Chucked is not reason for alarm. That it's actually an assertion failure.

Any thoughts of making the ChuckedAWobblyException a little more explicit in what it is and not a funny name?

jbogard:
...ay my bills). Closes #179


JakeGinnivan:
Nice, will get this released with a few of the other fixes which have been done recently.

Issue #190, opened 2015-03-23
Allow custom/additional message to ShouldlyMessage


aaronpowell:
I think this is related to #52 but figured I'd like to add it anyway.

I'd **really** like a way to provide a custom message to the 'assert' that is executed.

Scenario:

> I am doing a HTTP Request, I expect an `OK` response, I get `Internal Server Error`. I'd like to add to the message the server response.

Here's a sample of how I'd see the API working:

    [Fact]
    public async Task Should_return_OK() {
        var response = await GetAsync("/api/something");
        
        response.StatusCode.ShouldBe(HttpStatusCode.OK, () => response.Content.ReadAsStringAsync().Result);
    }

Notice the use of the lambda for the additional message, this prevents the code from being executed unless the Should fails, so if you're doing something complex or that has side effects (ie - reading the a HTTP response) it will only be done if needed.

The invocation of the lambda should be in a `try/catch` itself and if it throws just do a `Debug.WriteLine` with the exception (or something similar) and add something to the output message that the additional message failed.

fekberg:
I like the idea of having extra information in the ShouldlyMessage, just a small note on the asynchronous part where you are forcing a block. That code might deadlock, which isn't ideal. It's generally bad practice to force a wait/blocking.

You could change it in two ways.

**Async all the way**
`response.StatusCode.ShouldBe(HttpStatusCode.OK, async () => await response.Content.ReadAsStringAsync());`

**Force block and avoid deadlock**
This is a bit of a "hack" but will result in the wanted behaviour of blocking code. The reason you wrap it in a new Task is so that you are not locking the Task/Thread where the signal of the async completion will be notified to. In ASP.NET this also avoids capturing the `SynchronizationContext`.

`response.StatusCode.ShouldBe(HttpStatusCode.OK, () => Task.Run(() => response.Content.ReadAsStringAsync()).Result);`

robdmoore:
+1 I've wanted custom messages a couple of times. This example being one of them actually!

I'd want a string overload as well as the lambda though.

PureKrome:
Blammo! same here @aaronpowell and @robdmoore

pawelpabich:
+1 one for simple string overload

aaronpowell:
@fekberg just having a quick dig into the code, it supports .NET 3.5, would that have any problem with the async/await stuff?

fekberg:
[...]

PR #193, opened 2015-03-25
Added custom message support to ShouldBe

aaronpowell:
This is a WIP that starts addressing #190


JakeGinnivan:
@aaronpowell if you could retarget this to point at #195 then others can pitch in :)

PR #199, opened 2015-03-27
Added custom message to ShouldBeDateTimeTestExtensions.cs and ShouldBeNumericTestExtensions including tests

JoeMighty:
Here are the custom messages for `ShouldBeDateTimeTestExtensions.cs`. I hope it's all ok, I'm relatively new to Git!

### DIFF
--- a/src/Shouldly/ShouldlyExtensionMethods/ShouldBe/ShouldBeDateTimeTestExtensions.cs
+++ b/src/Shouldly/ShouldlyExtensionMethods/ShouldBe/ShouldBeDateTimeTestExtensions.cs
@@ -12,9 +12,29 @@ namespace Shouldly
             actual.AssertAwesomely(v => Is.Equal(v, expected, tolerance), actual, expected, tolerance);
         }
 
+        public static void ShouldBe(this DateTime actual, DateTime expected, TimeSpan tolerance, string customMessage)
+        {
+            actual.AssertAwesomely(v => Is.Equal(v, expected, tolerance), actual, expected, tolerance, () => customMessage);
+        }
+
+        public static void ShouldBe(this DateTime actual, DateTime expected, TimeSpan tolerance, Func<string> customMessage)
+        {
+            actual.AssertAwesomely(v => Is.Equal(v, expected, tolerance), actual, expected, tolerance, customMessage);
+        }
+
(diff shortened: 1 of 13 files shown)


src/Shouldly/ShouldlyExtensionMethods/ShouldBe/ShouldBeDateTimeTestExtensions.cs
```diff
@@ -12,29 +12,89 @@ public static void ShouldBe(this DateTime actual, DateTime expected, TimeSpan to
             actual.AssertAwesomely(v => Is.Equal(v, expected, tolerance), actual, expected, tolerance);
         }
 
+        public static void ShouldBe(this DateTime actual, DateTime expected, TimeSpan tolerance, string customMessage)
+        {
+            actual.AssertAwesomely(v => Is.Equal(v, expected, tolerance), actual, expected, tolerance, () => customMessage);
+        }
+
+        public static void ShouldBe(this DateTime actual, DateTime expected, TimeSpan tolerance, Func<string> customMessage)
```
JakeGinnivan: Could the other two overloads call this one, if we do internal refactoring later it just means less touch points if we call overloads rather than copy the code inside. i.e

```
 public static void ShouldBe(this DateTime actual, DateTime expected, TimeSpan tolerance, string customMessage)
{
    actual.ShouldBe(expected, tolerance, () => customMessage);
}
```

JoeMighty: Yes, of course. I can't believe I missed that.

JakeGinnivan: No worries, this is what code reviews are for :)

JoeMighty:
I've just added custom messages to `ShouldBeNumericTestExtensions.cs` and updated the tests too. 

Should I have added this as a separate pull request or is it ok to add it in this pull request?

JakeGinnivan:
It is fine to include it as it's related.

Generally if you want to split it up, just go back to master and take a new feature branch off master then you can have two branches working in parallel

JoeMighty:
Alright, will do! Thanks.

PR #200, opened 2015-03-28
Message generator for ShouldBeSubsetOf

### LINKED ISSUE #158: Message generator for ShouldBeSubsetOf
The error message is not very informative, it should say which items are additionally in the actual.

gabrielweyer:
I added a class called `ShouldBeSubsetOfMessageGenerator` to handle the formatting of the error message.

`ShouldBeSubsetOf<T>` now uses `Except` to determine which elements are missing.

Let me know if you want me to modify anything.

Fix #158

### DIFF
--- a/src/Shouldly/MessageGenerators/ShouldBeSubsetOfMessageGenerator.cs
+++ b/src/Shouldly/MessageGenerators/ShouldBeSubsetOfMessageGenerator.cs
@@ -1,4 +1,6 @@
-﻿using System.Text.RegularExpressions;
+﻿using System.Collections;
+using System.Linq;
+using System.Text.RegularExpressions;
 
 namespace Shouldly.MessageGenerators
 {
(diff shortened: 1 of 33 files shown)


src/Shouldly/MessageGenerators/ShouldBeSubsetOfMessageGenerator.cs
```diff
@@ -0,0 +1,24 @@
...
+        private static readonly Regex Validator = new Regex("ShouldBeSubsetOf", RegexOptions.Compiled);
+
+        public override bool CanProcess(ShouldlyAssertionContext context)
+        {
+            return Validator.IsMatch(context.ShouldMethod);
+        }
+
+        public override string GenerateErrorMessage(ShouldlyAssertionContext context)
+        {
+            const string format = @"{0} should be subset of {1} but {2} do not";
```
JakeGinnivan: Maybe one of the following instead (better English)

 - `{0} should be subset of {1} but {2} are outside subset`
 - `{0} should be subset of {1} but {2} are not`
 - `{0} should be subset of {1} but {2} are additional`

Not quite sure which is best. @robdmoore @xerxesb any thoughts?

gabrielweyer: Ah yes, sorry for my broken English :stuck_out_tongue: 

I like the third one `{0} should be subset of {1} but {2} are additional` the most.

JakeGinnivan: Maybe `{0} should be subset of {1} but {2} are outside subset` as it is quite explicit. 

Then this is getting really nitpicky but a single item should read `is` not `are`. For example

 - `new[] { 1, 2, 5 } should be subset of  [2, 3, 4] but [1, 5] are outside subset`
 - `new[] { 2, 3, 5 } should be subset of  [2, 3, 4] but [1] is outside subset`

gabrielweyer:
Let me know what you think.

JakeGinnivan:
Great! Thanks for persisting.

Something has gone funky with the PR so I rebased and merged manually at https://github.com/shouldly/shouldly/commit/59a303e301bdc89bc9678684262700f6ca0ea701

Suggest doing a fetch then `git reset --hard upstream/master` to clean up your master before doing anything else

PR #203, opened 2015-04-04
Detailed string diff

chaitanyagurrapu:
Initial implementation of detailed string diff. Fixes issue #41


chaitanyagurrapu:
New functionality will require update to the docs 

        [Test]
        public void ShouldBe()
        {
            var theSimpsonsCat = new Cat() { Name = "SnowBall2" };
            theSimpsonsCat.Name.ShouldBe("Snowball 2");
        }

Failure message

        theSimpsonsCat.Name
            should be
        "Snowball 2"
            but was
        "SnowBall2"
            difference
        Case Sensitive Comparison
    Difference     |                      |                   |    |   
                   |                     \|/                 \|/  \|/  
    Index          | 0    1    2    3    4    5    6    7    8    9    
    Expected Value | S    n    o    w    b    a    l    l    \s   2    
    Actual Value   | S    n    o    w    B    a    l    l    2         
    Expected Code  | 83   110  111  119  98   97   108  108  32   50   
    Actual Code    | 83   110  111  119  66   97   108  108  50        

There are further enhancements that can be made on top of this. Rather than doing them all as part of a large PR, I figured I would get the base functionality out first for review and then do the others as part of different issues/PRs

JakeGinnivan:
That is awesome. Few questions. How does this work for long strings?

NUnit does something like this:

```
Expected string length 3 but was 8. Strings differ at index 0.
  Expected: "Joe"
  But was:  "Chandler"
  -----------^
```

But with a "long" (>1000 symbols) string:
```
  Expected string length 1538 but was 1539. Strings differ at index 501.
  Expected: "...12211891925313013474130194418_first string_159148209974874..."
  But was:  "...12211891925313013474130194418_second string_15914820997487..."
  --------------------------------------------^
```

I also wonder if a downwards approach like at http://haacked.com/archive/2012/01/14/comparing-strings-in-unit-tests.aspx would work better? (implementation is at https://gist.github.com/Haacked/1610603)

chaitanyagurrapu:
I did start with the downward approach and used the code that was linked to in the original issue. But it was kind of hard to read English strings when they are laid out top to bottom. So I went this way instead.

Long strings is one of that things I figured we would deal as part of a different PR, if the current direction/code is approach is acceptable. 

Another enhancement would be to start using it in other string comparison methods such as "EndsWith", "BeginsWith" etc. Currently they don't have case sensitivity related option either. So we would have to add that functionality along with displaying differences.

After this is PR is merged, I can create issues for them and send different PRs for each of them. Or I can do it as part of this as well, whichever you would find easiest to review. :-)

JakeGinnivan:
I might create a separate branch to work on this feature so it is not merged until it is completed. It is not in a state where I can release the next version yet.

I'm just unsure if this is overkill. Maybe the NUnit way is simpler for the moment and we add either an overload or a global config option to opt into the full diff mode.

chaitanyagurrapu:
No worries. Another branch sounds good. Once the core functionality is bedded down, hopefully the truncation of long strings should be fairly straight forward. Although, there might still be some hidden complexity which might expand the amount of work required.

As for whether it is an overkill, It doesn't really seem like it to be honest. It seems very much the mandate of Shouldly to provide detailed, helpful messages. So a diff where the actual index number of the diff is displayed and the non printable characters are converted to printable characters seems very much like something Shouldly should do.

But yeah, the strings will have to be truncated lot shorter than 1000 characters though. Since we are displaying the index values, I was thinking it would be perhaps 10 or 15 characters wide.

Something like so ...

        theSimpsonsCat.Name
            should be
        "Start of a very long string which keeps on going and goin..."
            but was
        "Start of a very long string which keeps on going and goin..."
            difference
        Case Sensitive Comparison
    Difference     |                      |                   |      
                   |                     \|/                 \|/     
    Index          | ...  101  102  103  104  105  106  107  108  ... 
    Expected Value | ...  n    o    w    b    a    l    l    \s   ... 
    Actual Value   | ...  n    o    w    B    a    l    l    2    ... 
    Expected Code  | ...  110  111  119  98   97   108  108  32   ... 
    Actual Code    | ...  110  111  119  66   97   108  108  50   ...


    Difference     |            |                                
                   |           \|/                               
    Index          | ...  231  232  233  234  235  236  237  238  ... 
    Expected Value | ...  n    o    w    b    a    l    l    \s   ... 
    Actual Value   | ...  n    o    w    B    a    l    l    2    ... 
    Expected Code  | ...  110  111  119  98   97   108  108  32   ... 
    Actual Code    | ...  110  111  119  66   97   108  108  50   ...


    Difference     |                                     |      
                   |                                    \|/     
    Index          | ...  585  582  583  584  585  586  587  588  ... 
    Expected Value | ...  n    o    w    b    a    l    l    \s   ... 
    Actual Value   | ...  n    o    w    B    a    l    l    2    ... 
    Expected Code  | ...  110  111  119  98   97   108  108  32   ... 
    Actual Code    | ...  110  111  119  66   97   108  108  50   ...

Basically show all the differences in the strings at once along with a bit of context. Again, if there are too many differences, we might have to end up showing only the top 10 or something like that and indicating that there are more differences.

JakeGinnivan:
Replaced by  #205

JakeGinnivan:
Your idea looks great. Lets do that. I have added a TODO on the other PR and it has a branch in the main repo. We can target any further PR's to that.

Are there any other items you want to add to the TODO before we treat that PR as complete? Maybe put the comment on the other PR

PR #205, opened 2015-04-07
Detailed string diff

### LINKED ISSUE #41: Display detailed diff for string comparisons when calling ShouldBe

JakeGinnivan:
Fixes #41 

Todo:

- [ ] When more than 10 diffs in strings, show first 10 and indicate that there are more diffs


chaitanyagurrapu:
- [ ] When more than 10 diffs in strings, show first 10 and indicate that there are more diffs
- [ ] Implement detailed diffs for "Should[Not]StartWith" and "Should[Not]EndWith" assuming issue #206  has been resolved

JakeGinnivan:
Lets just get the long string issue sorted in this PR. The other task can be done separately

PR #229, opened 2015-04-20
ShouldContain and ShouldNotContain case sensitivity and tests

JoeMighty:


src/Shouldly.Tests/ShouldNotContain/StringContainsStringCaseIsInsensitiveScenario.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Shouldly.Tests.TestHelpers;
+
+namespace Shouldly.Tests.ShouldNotContain
+{
+    public class StringContainsStringCaseIsInsensitiveScenario : ShouldlyShouldTestScenario
+    {
+        protected override void ShouldThrowAWobbly()
+        {
+            "Shouldly is legendary".ShouldNotContain("legendary", Case.Insensitive);
```
JakeGinnivan: `.ShouldNotContain("LEGENDARY")`? Sensitive would also throw in this case?

JakeGinnivan:
Just one question added. Good addition

src/Shouldly.Tests/ShouldNotContain/StringContainsStringCaseIsInsensitiveScenario.cs
JoeMighty: No, `"Shouldly is legendary".ShouldNotContain("LEGENDARY", Case.Sensitive)` would pass (not throw a wobbly - causing the `ShouldThrowAWobbly` test to fail).

JoeMighty:
Not a problem, I've replied.

src/Shouldly.Tests/ShouldNotContain/StringContainsStringCaseIsInsensitiveScenario.cs
JakeGinnivan: But if we leave it as Insensitive?

JoeMighty: `"Shouldly is legendary".ShouldNotContain("LEGENDARY", Case.Insensitive)` throws an exception passing the WobblyTest (providing the wobbly error message is also updated to an upper-case "LEGENDARY" as the message thrown is `"Shouldly is legendary" should not contain "LEGENDARY" but does`.

I wonder whether it's worth updating the error message that's thrown to highlight that it's case insensitive to prevent possible confusion, or should that be left up to the user with the context that was recently added?

JakeGinnivan: Lets extend the error messages... We should also do it with ShouldBe I guess. https://github.com/shouldly/shouldly/blob/master/src/Shouldly.Tests/Strings/ShouldBe/CaseIsInsensitiveScenario.cs#L15

JoeMighty: Yeah, `ShouldBe` makes sense too. I'll sort it. I presume this would be done using a custom message generator?

JakeGinnivan: Yep I think so. I think it's better to do this all properly so we don't have to revisit

JoeMighty: I'm having a bit of difficulty in creating the ShouldNotContainCaseInsensitiveMessageGenerator. Within the `CanProcess` message I need to check the assertion context to see if the method signature is either Case.Sensitive or Case.Insensitive; however it doesn't look like this can be reflected from `MethodBase` returned by UnderlyingShouldMethod (presuming this is how I'm supposed to be going about it).

Do you have any suggestions?

JakeGinnivan: Hey,

Yeah you can create a new context type, like we do in https://github.com/shouldly/shouldly/blob/master/src/Shouldly/Internals/ShouldlyCoreExtensions.cs#L55 which holds the values properly

JakeGinnivan: Then you can also use

```
public override bool CanProcess(IShouldlyAssertionContext context)
{
    return context is ExpectedActualCaseInsensitiveShouldlyContext;
}
```

Rather than reflection

JoeMighty: I see, I wondered if that was the case!

Sorry, but I can't see a new context type being created on that line? I've searched my solution and can't see any classes that are derivatives of ShouldlyAssertionContext? Am I being blind?

JakeGinnivan:
Hey @JoeMighty thanks for the PR

If you are interested I did a little bit of follow up at https://github.com/shouldly/shouldly/commit/485c8c74e9b4ca62413b0015756bb16ef878531a
I just did some small bits of cleanup

JoeMighty:
I've just taken a look. It looks much better, thanks for sorting that. I was kind of on the right route then. You decided to put the caseSensitivity property into the context in the end?

JakeGinnivan:
There was an issue with another message generator generating the message before the new one. By putting on the context we could extend existing ones

Sent from my Windows Phone
________________________________

Issue #221, opened 2015-04-12
ShouldContainWithoutWhitespace message generator


JakeGinnivan:
Current:

```
Shouldly.ShouldAssertException : 
    ex.Message
        should contain without whitespace
    "Should not throw System.RankException but does
Additional Info:
Some additional context"
        but does not
```

It should have the actual message as well to make diagnosing the issue easier

urig:
How do I reproduce this?

urig:
Would it be acceptable to go:

    throw new ShouldAssertException(new ExpectedShouldlyMessage(ex, customMessage).ToString());

Instead of:

    throw new ShouldAssertException(new ExpectedShouldlyMessage(ex.GetType(), customMessage).ToString());

Here: https://github.com/shouldly/shouldly/blob/master/src/Shouldly/ShouldStaticClasses/ShouldThrowTaskExtensions.cs#L173 ? (and on line 169 naturally)

JakeGinnivan:
I explained it badly, it is not the Should.Throw methods, it is the `ShouldContainWithoutWhitespace` method. We use that method to verify others, I was writing tests for the Should.Throw method.

```
Shouldly.ShouldAssertException : 
    ex.Message
        should contain without whitespace
    "<expected>"
        but does not
```

Maybe something like:

```
Shouldly.ShouldAssertException : 
    ex.Message
        should contain without whitespace
    "<expected>"
        but
    "<actual>"
        does not
```

urig:
I'm afraid I still don't quite get what you're specifying here. Can we have a quick chat about it [here](https://gitter.im/urig/shouldly) sometime?

PR #230, opened 2015-04-20
Issues/221    should contain without whitespace message generator

urig:
For https://github.com/shouldly/shouldly/issues/221

### DIFF
--- a/src/Shouldly.Tests/ShouldContain/EmptyArrayScenario.cs
+++ b/src/Shouldly.Tests/ShouldContain/EmptyArrayScenario.cs
@@ -6,14 +6,15 @@ namespace Shouldly.Tests.ShouldContain
     {
         protected override void ShouldThrowAWobbly()
         {
-            new int[0].ShouldContain(1, "Some additional context");
+            var target = new int[0];
+            target.ShouldContain(1, "Some additional context");
         }
 
         protected override string ChuckedAWobblyErrorMessage
         {
             get
             {
-                return "new int[0] should contain 1 but was actually" +
+                return "target should contain 1 but was actually" +
                         "[]"  +
                        "Additional Info:" +
                        "Some additional context";
--- a/src/Shouldly.Tests/ShouldContain/EmptyArrayScenario.cs
+++ b/src/Shouldly.Tests/ShouldContain/EmptyArrayScenario.cs
@@ -13,7 +13,8 @@ namespace Shouldly.Tests.ShouldContain
         {
             get
             {
-                return "new int[0] should contain 1 but does not" +
+                return "new int[0] should contain 1 but was actually" +
+                        "[]"  +
                        "Additional Info:" +
                        "Some additional context";
             }
--- a/src/Shouldly.Tests/ShouldContain/LongStringScenario.cs
+++ b/src/Shouldly.Tests/ShouldContain/LongStringScenario.cs
@@ -4,21 +4,27 @@ namespace Shouldly.Tests.ShouldContain
 {
     public class LongStringScenario : ShouldlyShouldTestScenario
     {
+        protected string target = new string('a', 110) + "zzzz";
+
         protected override void ShouldThrowAWobbly()
         {
-            var longString = new string('a', 110) + "zzzz";
-            longString.ShouldContain("fff");
+            target.ShouldContain("fff");
         }
 
         protected override string ChuckedAWobblyErrorMessage
         {
-            get { return "longString should contain \"fff\" but was actually \"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\""; }
+            //
+            get
+            {
+                return "target should contain \"fff\" " +
+                       "but was actually \"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" +
+                       "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\"";
+            }
         }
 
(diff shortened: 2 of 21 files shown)


src/Shouldly.Tests/ShouldContain/LongStringScenario.cs
```diff
@@ -12,7 +12,7 @@ protected override void ShouldThrowAWobbly()
 
         protected override string ChuckedAWobblyErrorMessage
         {
-            get { return "longString should contain \"fff\" but does not"; }
+            get { return "longString should contain \"fff\" but was actually \"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\""; }
```
JakeGinnivan: We should have a max length on this. `But was "aaaaaaaaaaaaaaaaaaaaaaaaaaa....`. we should use the same length as elsewhere

src/Shouldly.Tests/ShouldContain/EmptyArrayScenario.cs
```diff
@@ -13,7 +13,8 @@ protected override string ChuckedAWobblyErrorMessage
         {
             get
             {
-                return "new int[0] should contain 1 but does not" +
```
JakeGinnivan: We should extract the actual into a variable, that way when looking at these exceptions it makes sense why the actual is printed again?

src/Shouldly/ChuckedAWobbly.cs
```diff
@@ -2,6 +2,7 @@
 
 namespace Shouldly
 {
+    [Obsolete]
```
JakeGinnivan: Maybe add a comment in to say it has been left for compability reasons. The other thing which makes me nervous about removing this when/if we next bump major is that it's internal, so people would be accessing via reflection if they are using it and will not receive build warnings.

JakeGinnivan:
Thanks @urig. there are just a few small things but looks great overall

urig:
I'll get on these and will ping you when done.

src/Shouldly.Tests/ShouldContain/LongStringScenario.cs
urig: Not really sure what you mean here. Can you elaborate? Maybe be open a dedicated issue and send it my way? Tx

src/Shouldly.Tests/ShouldContain/EmptyArrayScenario.cs
urig: @JakeGinnivan I've just updated my branch with this. Most tests have a new member called 'target' that is used in both passing and failing scenarios.

src/Shouldly.Tests/ShouldContain/LongStringScenario.cs
JakeGinnivan: Say we have:

```
var longstring = "@Shouldly is an assertion framework which focuses on giving great error messages when the assertion fails while being simple and terse.

This is the old Assert way:

Assert.That(contestant.Points, Is.EqualTo(1337));

For your troubles, you get this message, when it fails:

Expected 1337 but was 0

How it Should be:

contestant.Points.ShouldBe(1337);

Which is just syntax, so far, but check out the message when it fails:

contestant.Points should be 1337 but was 0

It might be easy to underestimate how useful this is. Another example, side by side:

Assert.That(map.IndexOfValue("boo"), Is.EqualTo(2));    // -> Expected 2 but was 1
map.IndexOfValue("boo").ShouldBe(2);                    // -> map.IndexOfValue("boo") should be 2 but was 1

Shouldly uses the code before the ShouldBe statement to report on errors, which makes diagnosing easier.

Read more about Shouldly and it's features at http://docs.shouldly-lib.net/";

longString.ShouldContain("aaaa"); // fails
```
That entire string will be dumped into the output, making it really hard to read. Having a max length on the string we put in the exception message will make it better

urig: So here: https://github.com/shouldly/shouldly/blob/master/src/Shouldly/ShouldlyExtensionMethods/ShouldBe/StringContainsTestExtensions.cs#L55 I can use `actual.clip(100)` just like here: https://github.com/shouldly/shouldly/blob/master/src/Shouldly/ShouldlyExtensionMethods/ShouldBe/StringContainsTestExtensions.cs#L40 ?

JakeGinnivan: Yep exactly

urig:
@JakeGinnivan Done: Now clipping ShouldNotContain 'actual' string at 100 chars (with trailing ellipsis). Re-enabled unit test for ShouldNotContain LongString scenario. Kindly review :)

src/Shouldly.Tests/ShouldContain/StringContainsStringScenario.cs
```diff
@@ -4,19 +4,21 @@ namespace Shouldly.Tests.ShouldContain
 {
     public class StringContainsStringScenario : ShouldlyShouldTestScenario
     {
+        protected string target = "Shouldly is legendary";
+
         protected override void ShouldThrowAWobbly()
         {
-            "Shouldly is legendary".ShouldContain("legend-wait for it-dary");
+            target.ShouldContain("legend-wait for it-ary");
```
JakeGinnivan: ha, nice catch

src/Shouldly/ChuckedAWobbly.cs
```diff
@@ -2,6 +2,7 @@
 
 namespace Shouldly
 {
+    [Obsolete("This class is only kept here for backwards compatibility. Please use ShouldAssertException instead.")]
```
JakeGinnivan: I think we can drop this, the class was internal so the only use would have been via reflection meaning no one but us will get build warnings

Issue #242, opened 2015-06-18
Dictionary Should/ShouldNot ContainKeyAndValue fails for null values


andymcdowall:
If you try and do an assertion against a dictionary key with a null value then Shouldly will throw a NullReferenceException:

```C#
var dictionary = new Dictionary<string,string>{ { "Key", null } };
dictionary.ShouldContainKeyAndValue("Key", "Value"); // will fail with a NullReferenceException
```

The culprits are the equality checks in `ShouldBeDictionaryTestExtensions`:
```C#
if(!dictionary[key].Equals(val))
```

andymcdowall:
Pull request #243 created for issue

PR #243, opened 2015-06-18
#242 Bugs/exception with null dictionary keys

andymcdowall:
This is a fix for issue #242 where the dictionary checks can throw NullReferenceExceptions.

I've changed the equality checks from doing `dictionary[key].Equals(value)` to a null-safe `Equals(dictionary[key],value)`. This fixes the main issue.

In addition to this I've altered the associated message generators such that they now report null values as `null` rather than `"null"`.


JakeGinnivan:
Thanks!

PR #246, opened 2015-06-21
Switching to AppVeyor for builds

JakeGinnivan:


gep13:
Thinking about it...

Do you actually need the matrix to complete this?  Can you not simply run the build twice, on the same machine, each targetting the correct version of the .Net Framework?

I would suggest that the use of some build script would be easier to achieve this.

JakeGinnivan:
Yeah, thats what I am thinking. Would be good to support multiple projects/steps in a single appveyor build, but for the moment will just do a simple build script which just calls msbuild twice

gep13:
You can do multiple steps just now in a yml file, you would just need to "own" some more of it, and not rely on AppVeyor calling MSBuild for you, i.e. make it your build step that calls into msbuild directly

JakeGinnivan:
How do you do that, I couldn't see that feature documented. Saw that for before/after build but not a build step

gep13:
Have a look here:

https://github.com/gep13/gep13.github.io/blob/source/appveyor.yml

``build_script``` section

gep13:
granted, that was for building an Octopress site. but don't see why the same principal couldn't be applied to your msbuild steps.

Issue #213, opened 2015-04-10
Add extension methods on System.Action for ShouldThrow etc


kpstolk:
It is easier in my opinion to have an extension on System.Action like this:

    Action action = () => { throw new NotImplementedException(); };
    action.Throws<NotImplementedException>();

instead of:

    Action action = () => { throw new NotImplementedException(); };
    Should.Throws<NotImplementedException>(action);

JakeGinnivan:
I think having these extension methods make sense. They should just proxy to the existing static methods.

All the Should.Throw tests should be moved over to use the new extension methods so all code paths are still tested

Happy to have a pull request for this

JakeGinnivan:
I think this is even more useful for the task overloads.

`await task.ShouldThrowAsync<InvalidOperationException>()`

JoeMighty:
kpstolks initial suggestion is to make the method `action.Throws<T>`, however I feel `action.ShouldThrow<T>` makes more sense as all other Shoudly methods start with the Should prefix. 

What are your thoughts?

JoeMighty:
Also, is there any need to pass the action in as an argument as well? Presumably we can just do:

`action.ShouldThrow<T>()` instead of `action.ShouldThrow<T>(action)`?

kpstolk:
No, my bad. I've updated the sample code.

JoeMighty:
Thanks for the clarification. That makes more sense :)

JakeGinnivan:
This will also have to be in addition to old API. This isn't worth a breaking change

JoeMighty:
I've been working on this one ([see comparison here](https://github.com/shouldly/shouldly/compare/master...JoeMighty:ShouldThrowExtensions)) and added the necessary extension methods and updated the tests accordingly; however I'm struggling updating the tests for the timeout scenarios `FuncOfTaskWithTimeoutScenario.cs` and `TaskWithTimeoutScenario.cs` that take an additional delegate as an argument.

Do you have any suggestions on how I can proceed with this, they seem quite specific tests that won't work with these updates?

JakeGinnivan:
Should be fine, the overload should be

`ShouldThrow<T>(this Task<T> task, Timeout timeout);`

Is that what you mean?

JoeMighty:
Yes, I realised I'd missed the Timeout overload methods :S I was also having difficulty in converting the tests that pass the task to a `Should.Throw(Action)` but figured it out in the end. Thanks.

[...]

PR #249, opened 2015-06-30
Added ShouldThrow and ShouldNotThrow extensions

JoeMighty:
Here are the ShouldThrow and ShouldNotThrow extensions from for issue #213. I've updated the tests accordingly - I hope they're all ok!

### DIFF
--- a/src/Shouldly/ShouldStaticClasses/ShouldThrowExtensions.cs
+++ b/src/Shouldly/ShouldStaticClasses/ShouldThrowExtensions.cs
@@ -8,6 +8,20 @@ namespace Shouldly
     [ShouldlyMethods]
     public static partial class Should
     {
+        /*** ShouldThrow(Action) ***/
+        public static TException ShouldThrow<TException>(this Action actual) where TException : Exception
+        {
+            return Throw<TException>(actual);
+        }
+        public static TException ShouldThrow<TException>(this Action actual, string customMessage) where TException : Exception
+        {
+            return Throw<TException>(actual, customMessage);
+        }
+        public static TException ShouldThrow<TException>(this Action actual, [InstantHandle] Func<string> customMessage) where TException : Exception
+        {
+            return Throw<TException>(actual, customMessage);
+        }
+        /*** Should.Throw(Action) ***/
         public static TException Throw<TException>([InstantHandle] Action actual) where TException : Exception
         {
             return Throw<TException>(actual, () => null);
@@ -34,6 +48,21 @@ namespace Shouldly
             throw new ShouldAssertException(new ExpectedShouldlyMessage(typeof(TException), customMessage).ToString());
         }
 
+        /*** ShouldNotThrow(Action) ***/
+        public static void ShouldNotThrow(this Action action)
+        {
+            NotThrow(action);
+        }
+        public static void ShouldNotThrow(this Action action, string customMessage)
+        {
+            NotThrow(action, customMessage);
+        }
+        public static void ShouldNotThrow(this Action action, [InstantHandle] Func<string> customMessage)
+        {
+            NotThrow(action, customMessage);
+        }
+
+        /*** Should.NotThrow(Action) ***/
         public static void NotThrow([InstantHandle] Action action)
         {
             NotThrow(action, () => null);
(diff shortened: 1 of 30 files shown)


src/Shouldly/ShouldStaticClasses/ShouldThrowExtensions.cs
```diff
@@ -8,6 +8,20 @@ namespace Shouldly
     [ShouldlyMethods]
     public static partial class Should
     {
+        /*** ShouldThrow(Action) ***/
+        public static TException ShouldThrow<TException>(this Action actual) where TException : Exception
```
JakeGinnivan: I think we should put these extensions in another class so we don't have `Should.ShouldThrow` and `Should.Throw` on the same class.

Maybe move it `ShouldThrowExensions` or something

JakeGinnivan:
Just the one comment. Looks awesome apart from that

JoeMighty:
Great! I'm just about to head off to bed now but I'll try move them into another class tomorrow during my lunch break.

JakeGinnivan:
:+1:

JoeMighty:
As `ShouldThrowExtensions.cs` already exists I moved the newly created overloads into classes respective Overload.cs classes (ie: `ShouldThrowExtensionsOverloads.cs`). I hope this is ok, if not then let me know and I'll update it.

For some reason it appears to have broken the AppVeyor build but I can't see why it would have done this. It's complaining about references to tasks?

JakeGinnivan:
Yeah, Shouldly is both .net 3.5 and .net 4.0. So you need the `#if net40` directtives around any code which references .net 4.0 stuff.

JakeGinnivan:
The solution opens as .net 4.0 only. you can run `msbuild shouldly.sln /p:targetframeworkversion=v3.5` to test locally

src/Shouldly/ShouldStaticClasses/ShouldThrowExtensionsOverloads.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using JetBrains.Annotations;
+
+namespace Shouldly
+{
+    public static partial class Should
```
JakeGinnivan: Change the class name as well, these methods still exist on the `Should` type.

JoeMighty: Good point, is this better:

    public static class ShouldThrowOverloads
    {
        public static TException ShouldThrow<TException>(this Action actual) where TException : Exception
        {
            return Should.Throw<TException>(actual);
        }
    ...

JakeGinnivan: yep, that would be great

JoeMighty: It looks like updating the extension to return `Should.Throw` will result in the error message changing from "action should throw" to "return Should throw". It'd be nice to keep the variable (`action`) in the message. Is it worth looking at a custom message generator or do you point me in the direction of another way around this problem that allows us to move these methods into their own class yet retain the helpful error message?

JakeGinnivan: Make sure you attribute the new class with 
```
[DebuggerStepThrough]
[ShouldlyMethods]
```

JakeGinnivan: Im heading to bed (back in Australia now). If you don't make progress let me know, I will merge then have a look at solving in the morning when I get up

JakeGinnivan: The PR looks great, i'm sure I can sort the small remaining issues

JakeGinnivan:
Have time this morning to do some OSS work, figured i'd get this merged as you have done all the hard work :) Thanks!

JoeMighty:
Alright, thanks for picking that up for me.

JakeGinnivan:
Not a prob, was hardly anything to do didn't seem worth another round trip :+1:

PR #258, opened 2015-07-21
Added ShouldBeTrue|False and Null|NotNull extensions.

JoeMighty:
Added True/False extensions.


JoeMighty:
I'm just about finished on the Null|NotNull extensions which I'll merge into this pull request.

JoeMighty:
The Null|NotNull extensions are now included within this PR. I hope they're all ok!

PR #259, opened 2015-08-12
Should not throw message generator

### LINKED ISSUE #220: Should.NotThrow message generator
```
Shouldly.ShouldAssertException : 
    Should
        error 
    Shouldly.ShouldAssertException
        but does not
```

Is not great. Something like this would be better:

```
Shouldly.ShouldAssertException : 
   {code argument}
       should not throw but threw
   {exception type}
       with message
   {exception message}
```

yannisgu:
This PR will fix #220

I took the branch from @JoeMighty and started to implement the code to extract the expression passed to `Should.NotThrow`.  ~~At the moment in only works with the `Should.NotThrow(action, customMessage)` overload and only if action doesn't contain a comma (,). Also the tests proposed by @JakeGinnivan in #220 needs to be added.~~

The current code to extract the parameter counts the open and closing parenthesis and search for `,` or `)` to find the end of the parameter. I added some tests proposed by @JakeGinnivan but not all of them. Should I add more?

### DIFF
--- a/ShouldlyConvention.Tests/ShouldlyConvention.Tests.csproj
+++ b/ShouldlyConvention.Tests/ShouldlyConvention.Tests.csproj
@@ -101,6 +101,9 @@
       <Name>Shouldly</Name>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
+  </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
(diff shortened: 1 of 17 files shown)


JoeMighty:
Thanks for picking this one up and I'm glad what I've done is of help. I'm interested to see how extracting the arguments is done as that's the part I quickly got stuck on.

yannisgu:
Updated the current state in the PR description. From my point of view, ready for a review.

JakeGinnivan:
I have rebased this locally, tested and made it work for Should.Throw as well.

Nice work you two, this is an awesome addition. Will merge tomorrow

JoeMighty:
Cool! I did the easy bit, props to Yannisgu for tackling the hard part!

PR #261, opened 2015-08-20
Should not throw message generator

JakeGinnivan:


JakeGinnivan:
@JoeMighty @yannisgu went a bit further on the original PR

Can one of you review?

PR #317, opened 2015-10-02
ShouldNotBe error message PR

cmatskas:


src/Shouldly/ShouldlyMessage.cs
```diff
@@ -240,20 +240,21 @@ public string CreateExpectedErrorMessage()
...
-    {3}",
-                codePart, context.ShouldMethod.PascalToSpaced(), context.Expected.ToStringAwesomely(), context.Actual.ToStringAwesomely());
+            var message = string.Format(@"
+            {0}
+                {1}
+            {2}
+                but was
+            {3}",
+                codePart, context.ShouldMethod.PascalToSpaced(), context.Expected.ToStringAwesomely(),
+                context.ShouldMethod == "ShouldNotBe" ? string.Empty : context.Actual.ToStringAwesomely());
```
JoeMighty: I wonder whether it's worth moving this into a custom message generator? Thoughts @JakeGinnivan?

JakeGinnivan: Not sure it's worth splitting it, but we can use `context.IsNegatedAssertion` here instead

src/Shouldly/ShouldlyMessage.cs
```diff
@@ -240,20 +240,21 @@ public string CreateExpectedErrorMessage()
...
             var codePart = context.CodePart;
-            string message = string.Format(@"
-    {0}
-        {1}
-    {2}
-        but was
-    {3}",
-                codePart, context.ShouldMethod.PascalToSpaced(), context.Expected.ToStringAwesomely(), context.Actual.ToStringAwesomely());
+            var message = string.Format(@"
+            {0}
```
JakeGinnivan: Please fix the indenting of the string. This will put a heap of extra whitespace into the error messages.

src/Shouldly/ShouldlyMessage.cs
```diff
@@ -225,10 +225,10 @@ private string GenerateShouldMessage()
         public string CreateExpectedErrorMessage()
         {
             var format = @"
-    {0}
-        {1} {2}
-    {3}
-        but does{4}";
+            {0}
```
JakeGinnivan: Ditto, please correct the indenting

src/Shouldly/ShouldlyMessage.cs
```diff
@@ -240,20 +240,21 @@ public string CreateExpectedErrorMessage()
...
+                codePart, context.ShouldMethod.PascalToSpaced(), context.Expected.ToStringAwesomely(),
+                context.ShouldMethod == "ShouldNotBe" ? string.Empty : context.Actual.ToStringAwesomely());
 
             if (DifferenceHighlighter.CanHighlightDifferences(context))
             {
                 message += string.Format(@"
-        difference
-    {0}",
-                DifferenceHighlighter.HighlightDifferences(context));
+                difference
```
JakeGinnivan: Same here, fix indent

cmatskas: Morning guys, thanks for the feedback. Would it not be better if we used a StringBuilder instead to insert new lines and tabs where needed rather than relying on hardcoded white space?
Something along the lines of:

    var messageSb = new StringBuilder();
            messageSb.AppendLine(codePart);
            messageSb.AppendLine(string.Format("\t{0}", context.ShouldMethod.PascalToSpaced()));
            messageSb.AppendLine(context.Expected.ToStringAwesomely());
            messageSb.AppendLine("\tbut was");
            
            if (context.ShouldMethod != "ShouldNotBe")
            {
                messageSb.AppendLine(context.Actual.ToStringAwesomely());
            }

            if (DifferenceHighlighter.CanHighlightDifferences(context))
            {
                messageSb.AppendLine("\tdifference");
                messageSb.AppendLine(DifferenceHighlighter.HighlightDifferences(context));
            }

            return messageSb.ToString();

JakeGinnivan: Not as readable from a source point of view. Editors should not adjust these indents automatically as that is actually changing code?

cmatskas: @JakeGinnivan thanks for the response. I think Resharper was overly keen to fix indentation. I should have known better. I will revert the space changes and push that in. Thanks

JakeGinnivan: Hrrm, if you can reproduce that we should raise a bug. It shouldn't change the indentation of that..

cmatskas: Ha! I thought it was R#. I need more coffee! No bug here, nothing to see, move along :)

cmatskas:
Hey folks, any progress on this?

JoeMighty:
I think the only thing that's left is Jake's comment about using `context.IsNegatedAssertion`. Other than that it looks good to me.

src/Shouldly/ShouldlyMessage.cs
cmatskas: I'm not sure that Context.IsNegatedAssertion would work here. It returns true if the shouldMethod contains "Not" but we explicitly want to check for "ShouldNotBe". Maybe excent the context with a isShouldNotBe assertion? At some point, the explicit check will have to be done using a hard coded string, unless we introduce some kind of enum. Thoughts?

cmatskas:
OK, we've decided to use a new ShouldBeMessageGenerator to handle both positive and negative assertions. I'll get on with it

Issue #312, opened 2015-09-29
ShouldNotMatch method


JakeGinnivan:
For matching regex's, the inverse overload is missing. See `ShouldMatch`

bradarv90:
I would like to look into making this.

JoeMighty:
This has now been commited, thanks for all of your help with it @bradarv90

PR #319, opened 2015-10-15
Made ShouldNotMatch with a test

bradarv90:
For #312

### DIFF
--- a/src/Shouldly.Tests/Strings/ShouldNotMatch.cs
+++ b/src/Shouldly.Tests/Strings/ShouldNotMatch.cs
@@ -2,7 +2,7 @@
 
 namespace Shouldly.Tests.Strings
 {
-    class ShouldNotMatch : ShouldlyShouldTestScenario
+    public class ShouldNotMatch : ShouldlyShouldTestScenario
     {
         protected override void ShouldThrowAWobbly()
         {
--- a/src/Shouldly.Tests/Strings/ShouldNotMatch.cs
+++ b/src/Shouldly.Tests/Strings/ShouldNotMatch.cs
@@ -1,0 +1,27 @@
+﻿using Shouldly.Tests.TestHelpers;
+
+namespace Shouldly.Tests.Strings
+{
+    class ShouldNotMatch : ShouldlyShouldTestScenario
+    {
+        protected override void ShouldThrowAWobbly()
+        {
+            "Cheese".ShouldNotMatch(@"\w+", "Some additional context");
+        }
+
+        protected override string ChuckedAWobblyErrorMessage
+        {
+            get
+            {
+                return "\"Cheese\" should not match \"\\w+\" but was \"Cheese\" " +
+                       "Additional Info: " +
+                       "Some additional context";
+            }
+        }
+
+        protected override void ShouldPass()
+        {
+            "Cheese".ShouldNotMatch(@"Cat");
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


bradarv90:
I'm not sure what this failed tests means. Is that related to my code or just a flakey test?

src/Shouldly.Tests/Strings/ShouldNotMatch.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Shouldly.Tests.TestHelpers;
+
+namespace Shouldly.Tests.Strings
+{
+    class ShouldNotMatch : ShouldlyShouldTestScenario
```
JoeMighty: Looks good so far, don't forget to mark the class as public.

JoeMighty:
No, it's not you - it appears to be related to an API approval library we use. I'll try and have a look into it.

bradarv90:
Ok. I made the class public. Thanks

JakeGinnivan:
The API approval library is there so all public API changes have to be *approved*. This stops accidental breaking changes.

You should be able to run the test and a diff tool will pop up. You can then merge the new apis over into the .approved.cs file

src/Shouldly.Tests/Strings/ShouldNotMatch.cs
```diff
@@ -0,0 +1,27 @@
...
+        protected override void ShouldThrowAWobbly()
+        {
+            "Cheese".ShouldNotMatch(@"\w+", "Some additional context");
+        }
+
+        protected override string ChuckedAWobblyErrorMessage
+        {
+            get
+            {
+                return "\"Cheese\" should not match \"\\w+\" but was \"Cheese\" " +
```
JakeGinnivan: Similar to the `ShouldNotBe`, if this test fails the `but was...` will always repeat the expected.

I would just make the error `"\"Cheese\" should not match \"\\w+\" but did`

JoeMighty:
That API Approval tool is awesome. I got the diff tool popup as you described - it's pretty slick isn't it!.

src/Shouldly.Tests/Strings/ShouldNotMatch.cs
bradarv90: I'm struggling to understand how ShouldNotBe changes the normal error message. Would you mind explaining?

JakeGinnivan: Hey, sorry for the delay @bradarv90 

an example of the message generators is https://github.com/shouldly/shouldly/blob/master/src/Shouldly/MessageGenerators/ShouldBeMessageGenerator.cs

You could create a ShouldMatchMessageGenerator and wire it up in the same way, that then lets you customise the messages

JoeMighty: How are you getting on with this @bradarv90? If you need any help or don't have the time to finish it off then I'm happy to finish the last bit off.

bradarv90: I'm sorry. I've gotten so busy with school work. I think it would be best if you finished it.

JoeMighty: No problem! Thanks for all you've done :)

JoeMighty:
As @bradarv90 wasn't able to finish this one off I finished it as part of pr329 by adding the ShouldNotMatchMessageGenerator.

PR #323, opened 2015-10-26
fixed whitespace

adamralph:
Sorry just had to do that before continuing with my keynote at .NET Unboxed!


JakeGinnivan:
Thanks!

Issue #327, opened 2015-11-09
ShouldThrow fails under Mono


nbarbettini:
Running this test code with the latest build of Mono:

```csharp
var task = Task.Factory.StartNew(() => { throw new ArgumentNullException(); },
	CancellationToken.None, TaskCreationOptions.None,
	TaskScheduler.Default);

Should.Throw<ArgumentNullException> (async () =>
{
	await task;
});
```

Throws this error message:

> The provided expression` should throw System.ArgumentNullException but threw System.NullReferenceException

JoeMighty:
This has now been fixed, thanks for the contribution @nbarbettini!

PR #328, opened 2015-11-09
Fix for #327 Mono NullReferenceException

nbarbettini:
`preserveStackTrace` is null when running under Mono, so `.Invoke` fails. After this fix, `Should.Throw<>` behaves as expected.


JoeMighty:
Thanks a lot for the fix!

nbarbettini:
No prob! :+1:

PR #322, opened 2015-10-24
[WIP] #302 Add line ending insensitive assertions for strings

sholland1:

Issue #302, opened 2015-09-29
string.ShouldBeIgnoringLineEndings(string)


JakeGinnivan:
Line endings can sometimes be a pain, environmental factors can mean sometimes \r\n is used, others just \n. When doing string assertions we should be able to just ignore line endings.

sholland1:
Sounds easy enough. I will get started on this.

sholland1:
Two things:
1 I'm assuming you mean ignoring the *type* of line ending. In other words
```
"line1\r\nline2".ShouldBeIgnoringLineEndings("line1\nline2") //should pass
"line1\r\nline2".ShouldBeIgnoringLineEndings("line1line2") //should fail
```
2 Ideally, ignoring line endings would be an option similar to how case sensitivity is now.
I'm thinking the ideal API would be:
```
public static void ShouldBe(
    this string actual,
    string expected,
    bool ignoreCase = false,
    bool ignoreLineEndings = false)
{
    //impl
}
public static void ShouldBe(
    this string actual,
    string expected,
    string customMessage = null,
    bool ignoreCase = false,
    bool ignoreLineEndings = false)
{
    //impl
}
public static void ShouldBe(
    this string actual,
    string expected,
    Func<string> customMessage = null,
    bool ignoreCase = false,
    bool ignoreLineEndings = false)
{
    customMessage = customMessage ?? () => null;
    //impl
}
```
, but this would obviously be a breaking change.

So, do you have any thoughts on what the API should be for this?

JakeGinnivan:
I think we should introduce a flags enum to ShouldBe. Then you can add an obsolete attribute to the current one.

``` csharp
[Obsolete("Use the ShouldBeStringOptions instead of the ignoreCase boolean flag")]
public static void ShouldBe(this string actual, string expected, string customMessage, bool ignoreCase)
{
    actual.ShouldBe(expected, customMessage, ShouldBeStringOptions.IgnoreCase);
}
[Obsolete("Use the ShouldBeStringOptions instead of the ignoreCase boolean flag")]
public static void ShouldBe(this string actual, string expected, Func<string> customMessage, bool ignoreCase)
{
    actual.ShouldBe(expected, customMessage, ShouldBeStringOptions.IgnoreCase);
}
public static void ShouldBe(
    this string actual,
    string expected,
    ShouldBeStringOptions options = ShouldBeStringOptions.NormaliseLineEndings)
{
    //impl
}
public static void ShouldBe(
    this string actual,
[...]

PR #333, opened 2015-11-19
Should be with options

JakeGinnivan:
Superseeds #322 

@sholland1 and @JoeMighty, thoughts on the changes I have made over the original?


sholland1:
I was about to make some suggestions...

JakeGinnivan:
I am putting up another xUnit PR based on this, the diff was really big so merged this.

Feel free to make suggestions still and we can refine before release

sholland1:
Ok, cool. I should have some time to work on it tomorrow.
Should I comment on the new PR or this one?

src/Shouldly/Internals/Assertions/StringShouldBeAssertion.cs
```diff
@@ -0,0 +1,55 @@
+﻿using Shouldly.DifferenceHighlighting2;
+using System;
+
+namespace Shouldly.Internals.Assertions
+{
+    internal class StringShouldBeAssertion : IAssertion
```
JakeGinnivan: Just looking into this a bit more with the xunit changes. Why not use the message generator patterns?

sholland1: I pretty much explain my thinking here: https://github.com/shouldly/shouldly/issues/302#issuecomment-149386251

PR #335, opened 2015-11-21
Xunit conversion

JakeGinnivan:
Other fixes which will be included:

 - Improved Dictionary related exception messages
 - Better error messages overall when no source is available


src/Shouldly.Tests/Strings/ShouldBeNullOrEmpty.cs
```diff
@@ -1,24 +1,47 @@
...
 {
-    public class ShouldBeNullOrEmpty : ShouldlyShouldTestScenario
+    public class ShouldBeNullOrEmpty
     {
-        protected override void ShouldThrowAWobbly()
+        [Fact]
+        public void SingleLetterShouldFail()
         {
-            "a".ShouldBeNullOrEmpty("Some additional context");
+            Verify.ShouldFail(() =>
```
JakeGinnivan: @JoeMighty this is what I am thinking for tests going forward.

It tests the error messages with and without source being accessible. Thoughts?

JakeGinnivan:
@sholland1 if you are interested in helping out, am looking for thoughts on this change in direction

sholland1:
I don't have experience with xUnit, so I wouldn't be much help on that front.

src/Shouldly.Tests/InternalTests/DifferenceIndexConsolidatorTests.cs
```diff
@@ -1,16 +1,16 @@
...
 {
-    [TestFixture]
     public class DifferenceIndexConsolidatorTests
     {
         const int MaxDiffLength = 5;
         const int MaxLengthOfStrings = 21;
 
-        [Test, TestCaseSource("IndexConsolidationTestCases")]
+        [Theory]
+        [PropertyData("IndexConsolidationTestCases")]
```
AmadeusW: Use `Xunit.MemberData` instead of obsolete `Xunit.Extensions.PropertyData` [more info](https://xunit.github.io/docs/test-migration.html)
The former is in Xunit v2, and the latter in Xunit v1. Also, there is no need to use `Xunit.Extensions` namespace in this situation.

sholland1:
While we are looking at improving tests, I think we should refactor some of the tests so that they integrate better with the Visual Studio test runner.

The following screenshot shows tests that are easy to work with:
good_tests
The name of the tests shows up in the Visual Studio test runner, and you can double click the tests to go straight to it.

The following screenshot shows tests that are a pain to work with:
bad_tests
I have to hover over the test to find out what it actually is. Then, to find the test's code, I do a search either using Ctrl+F or in the Solution Explorer window.

src/Shouldly.Tests/InternalTests/DifferenceIndexConsolidatorTests.cs
JakeGinnivan: This is something I am not done with, I am using the RC of the new NuGet package manager and I can't install xunit 2.0... I also wrote most of this on a plane so I manually shoehorned it in. I will fix properly and use xunit 2.0 before I complete this PR

JakeGinnivan:
@sholland1 yep, the new tests report nicely. I am aware of this and it's one of the motivations.

sholland1:
I see. Awesome!

JakeGinnivan:
FYI @JosephWoodward, I fixed a few detailed string diff tests in  ea96131

The main changes:

 - Remove 4 space indent
 - Line everything up in the detailed string diff (the second line has an extra space in nearly all tests)
 - Fill end of lines so they match (first two lines are missing spaces to make line length the same as the others)
 - Additional info has one blank line instead of 2 now
 - Escape characters do not work in multiline string literals. So have got `""StringOne" + "\a" + @"BackslashA""`

Then I made sure strings were pulled out to variables. Hope that helps if you get to doing some more tests

JosephWoodward:
Alright, thanks for the heads up @JakeGinnivan. Looks good to me.

JakeGinnivan:
@sholland1 hey, can you have a look at 842f19d

I got quite confused with the fact there was two string difference highlighters and one looked like it was not being used by the tests so I got rid of it, the messages seem fine without it. Let me know if you see an issue with the change.

sholland1:
About the second string difference highlighter, I added the second one as I was implementing line ending insensitive comparison, and I just didn't realize that I eliminated the need for the original.
So, yeah, this looks good.

JakeGinnivan:
@JosephWoodward so close now, think all the tests are fixed. Need to do a once over on the tests to make sure no issues are introduced (I have changed a lot of message generators) and fix whatever the build error is. I hopefully will get a chance after work tomorrow (after 6pm +8)

PR #336, opened 2015-11-22
String options clean up

sholland1:


src/Shouldly/Internals/SourceCodeTextGetter.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Shouldly.Internals
 {
-    internal class ActualCodeTextGetter : ICodeTextGetter
+    internal class SourceCodeTextGetter : ICodeTextGetter
```
JakeGinnivan: Nice rename

JakeGinnivan:
Nice changes

src/Shouldly/Internals/SourceCodeTextGetter.cs
sholland1: Thanks. :+1:

PR #338, opened 2015-11-25
Minor consistency tweaks to StringHelper.cs

JoeMighty:
Few minor consistency tweaks whilst browsing over `StringHelper.cs` during my lunch break.


src/Shouldly/Internals/StringHelpers.cs
```diff
@@ -68,7 +66,7 @@ internal static string ToStringAwesomely(this object value)
             }
 #endif
 
-            return value == null ? "null" : value.ToString();
+            return value.ToString();
```
JoeMighty: value is already checked for this condition at the start of the method.

PR #341, opened 2015-12-30
Add support for coreclr/portable libraries

JakeGinnivan:
Using shared projects and individual projects for each target, same for test projects. Make it easier to get IDE support for all different targets


JakeGinnivan:
@jbogard @onovotny can you guys give this a once over, I wanted to merge so I can sort docs out now but let me know if you have any suggestions on how to set this up better?

onovotny:
At a high level, I think it looks ok, but it seems to be missing `dotnet` support that lets DNX/CoreCLR work correctly.

The lib that's going into `\lib\portable-` needs to also `\lib\dotnet` and then you should add the `NuSpec.ReferenceGenerator` package to your portable project and configure it to update the `dotnet` dependency group for references.

PR #345, opened 2015-12-31
Created documentation samples project which outputs files to be inclu…

JakeGinnivan:
…ded in docs

 - Uses roslyn to parse the source to get the correct bits
 - Uses ShouldMatchApproved to output the code and exception text
 - Because the snippets are their own files and are created via tests we can refactor/update things and documation samples will always be up to date!


JakeGinnivan:
image
image
image
image
image

This should make the docs really easy to port over

JosephWoodward:
"Because the snippets are their own files and are created via tests we can refactor/update things and documation samples will always be up to date!" Wow, this sounds amazing - good job! :+1:

daveaglick:
I like the approach of calling the documentation generator directly from the tests and using the call stack to find the correct source file (and method) for Roslyn parsing - very clever.

One thing I did notice is that you're going to end up having Roslyn reparse the same source file for each test - might not be a big deal for a small number of tests per file, but could start to cause some slowness for large test files (though even then, may not be a big deal). In any case, a simple optimization would be to cache the syntax tree by source file name to avoid the parsing step when you've already seen it.

JakeGinnivan:
@daveaglick nice suggestion. Done in ec2dbab

JakeGinnivan:
FYI @enkafan 

Thanks for your initial work on getting sphinx setup

enkafan:
this is very cool. I love it

PR #349, opened 2016-01-02
Initial attempt at getting 'DynamicShould' working again.

chaitanyagurrapu:
Re visiting issue #145 again.


JakeGinnivan:
I have submitted a PR to your branch fixing a bug which means both the with and without source are using source.

chaitanyagurrapu:
I've updated the PR so that shouldly itself doesn't throw if there is no stacktrace/source code.

JakeGinnivan:
Thanks heaps for getting this working again :)

If you have time docs for DynamicShould would be amazing. All good if not

JakeGinnivan:
Rebased and put up #358 instead of this

chaitanyagurrapu:
@JakeGinnivan  no worries. Will submit a PR with document updates soon.

PR #351, opened 2016-01-08
Added ShouldBeNullNotNull docs

JosephWoodward:


JosephWoodward:
Just noticed the example `.cs` file doesn't match the same convention as the ShouldBeTrue/False. I'm happy to fix this before accepting the pull request if it's an issue? If not I'll sort it on the next commit.

Issue #326, opened 2015-11-08
ShouldContain(Predicate) with count


AmadeusW:
I would like to test that a collection contains a specific number of elements specified by a predicate.
My use case:
a collection with elements `"a", "b", "c", "c"`
```csharp
collection.Count().ShouldBe(4);
collection.ShouldContain(n => n.ToString().Contains("a"));
collection.ShouldContain(n => n.ToString().Contains("b"));
collection.ShouldContain(n => n.ToString().Contains("c")); // Check that it contains two elements with "c"
```
Possible implementations:
```csharp
collection.ShouldContain(n => n.ToString().Contains("c"), 2); // overload
collection.ShouldContainCount(n => n.ToString().Contains("c"), 2); // another method
```
Is this something you would like to see in shouldly, or is it too specific? I'd be happy to contribute if you're ok with this.

JakeGinnivan:
Sounds good to me, happy to get a PR for this

chaitanyagurrapu:
One of the big advantages of shouldly and other assertion frameworks is that it is meant to make the tests easier to read and debug. I am not sure this increases clarity of the assertions though. 

If I was reading code, the following statement

    collection.where(n => n == "c").Count().Shouldbe(2)

is very expressive and self evident even if I had never come across shouldly. Not so much either of the following statements 

    collection.ShouldContain(n => n.ToString().Contains("c"), 2); // overload 
    collection.ShouldContainCount(n => n.ToString().Contains("c"), 2); // another method

The number 2 is just hanging there. I would have to mouse over or look up shouldly or at the least read  these lines a few times before I can guess what the multiple 'counts' and 'contains' in a single line are referring to.

I don't mean to disparage the idea. This seems like a commonly used scenario that could possibly be improved. But unless we can find a better and clearer way of expressing the intent than we currently have, adding more specialized assertions makes the situation worse, not better. 

Just my 2 cents :-)

urig:
Please pardon my directness, but why is this issue still open? What's the action item at this time?

PR #352, opened 2016-01-10
for issue #326, added ShouldContainMatchingCount

mgroves:
I'm sure the unit tests and/or test messages could be improved, but I think commit will give you the Should extension that you want. This is for issue #326


src/Shouldly.Shared/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
```diff
@@ -47,6 +47,15 @@ public static void ShouldContain<T>(this IEnumerable<T> actual, [InstantHandle]
         {
             ShouldContain(actual, elementPredicate, () => null);
         }
+        public static void ShouldContainMatchingCount<T>(this IEnumerable<T> actual, [InstantHandle] Expression<Func<T, bool>> elementPredicate, int expectedCount)
```
JakeGinnivan: I think this should be an overload of `ShouldContain` for discovery?

src/Shouldly.Tests.Shared/ShouldContain/CollectionPredicateWithCountScenario.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        [Fact]
+        public void CollectionWithACountOtherThanTheSpecifiedCountOfMatchingPredicatesFails()
+        {
+            var collection = new[] { "a", "b", "c", "c" };
+            Verify.ShouldFail(() =>
+            collection.ShouldContainMatchingCount(x => x == "c", 5),
+
+errorWithSource:
+@"[""a"", ""b"", ""c"", ""c""]
```
JakeGinnivan: With source it should be `collection` at the start?

JakeGinnivan:
Make sure you are running the tests under debug build, not release

JakeGinnivan:
Thanks for the PR :)

mgroves:
My bad; some sort of local snafu in saving a file. New commit on the way.

mgroves:
Well I sure am making a mess. Now I'm failing convention tests. I'll work on it; sorry for the noise :)

JakeGinnivan:
Not a problem, am working on GitVersion atm so am around if you have any questions or need a hand

src/Shouldly.Shared/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
mgroves: Whoops, didn't see this comment. I'll rename; is this also causing a test failure?

JakeGinnivan:
There is a test which approves the public API. You just have to run that test manually and it will pop-up a difftool which you merge your API changes in.

The reason for this is so it makes it super easy to review public API changes (so we don't accidentally make breaking changes)

mgroves:
Ah, interesting, I've not seen that before.

JakeGinnivan:
Latest Shouldly feature includes the feature - http://shouldly.readthedocs.org/en/latest/assertions/shouldMatchApproved.html :)

mgroves:
There we go. Thanks for helping me; I hope this allows you to close out an issue, and maybe @AmadeusW will get some use out of it :)

src/Shouldly.Shared/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
```diff
@@ -48,6 +48,26 @@ public static void ShouldContain<T>(this IEnumerable<T> actual, [InstantHandle]
...
+        {
+            var condition = elementPredicate.Compile();
+            var actualCount = actual.Count(condition);
+            if (actualCount != expectedCount)
+            {
+                throw new ShouldAssertException(new ExpectedActualShouldlyMessage(expectedCount, actual, () => null).ToString());
+            }
+        }
+
+        public static void ShouldContain<T>(this IEnumerable<T> actual, [InstantHandle] Expression<Func<T, bool>> elementPredicate, int expectedCount, string customMessage)
```
JakeGinnivan: I think the `Func<string> customMessage` overload is missing.

Also you can make each of these methods call each other.

src/Shouldly.Shared/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
```diff
@@ -48,6 +48,26 @@ public static void ShouldContain<T>(this IEnumerable<T> actual, [InstantHandle]
...
+        {
+            var condition = elementPredicate.Compile();
+            var actualCount = actual.Count(condition);
+            if (actualCount != expectedCount)
```
JakeGinnivan: I think the logic is wrong for this, I believe the overload should be counting the number of items matched by the predicate.

as an example

var arr = new[] { 1, 2, 3, 4 };
arr.ShouldContain(i => i %2, 2); // Should have exactly 2 items which matches predicate

src/Shouldly.Shared/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
mgroves: Yep, that makes sense. Added that in another commit.

src/Shouldly.Shared/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
mgroves: I think the logic is correct. I added your example in the unit test (although I assumed that you meant:

```
arr.ShouldContain(i => (i %2) == 0, 2);
```

i.e. a predicate to get even numbers).

JakeGinnivan:
There are a few more things (which will also make the error message awesome).

When creating the exception, do this:

```
// Create the type in ShouldlyMessage.cs
internal class ShouldContainWithCountShouldlyMessage : ShouldlyMessage
{
    public ShouldContainWithCountShouldlyMessage(object expected, object actual, int matchCount, [InstantHandle] Func<string> customMessage, [CallerMemberName] string shouldlyMethod = null)
    {
        ShouldlyAssertionContext = new ShouldlyAssertionContext(shouldlyMethod, expected, actual)
        {
            HasRelevantActual = true,
            MatchCount = matchCount
        };
        if (customMessage != null) ShouldlyAssertionContext.CustomMessage = customMessage();
    }
}

// Throw this instead
throw new ShouldAssertException(new ShouldContainWithCountShouldlyMessage(elementPredicate.Body, actual, expectedCount, customMessage).ToString());

// Finally, update the ShouldContainPredicateMessageGenerator to use MatchCount if set in the error message
```

Also, would you be able to update the docs for this guy :P I am not sure if ShouldContain has been pulled across yet, but if you are up for migrating those docs then adding this then that wraps this up entirely!

src/Shouldly.Shared/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
JakeGinnivan: You are right, the error message confused me. Have put another comment how to fix the error message :)

mgroves:
Geez I thought this was going to be easy :)

Just committed the improved error messaging. I'll take a look at docs

JakeGinnivan:
Yeah, sorry about that. I am heading out for a few hours. Will get back to this once back

JakeGinnivan:
But changes are looking good!

mgroves:
That latest build fail seems to be choking on "x = c" vs "x == c". I don't know it would be generating single vs double equals? Seems to be passing locally just fine. Any ideas?

It looks like docs are using "Sphinx" or something? I've not heard of it, but I'm willing to give it a shot if I have time, but it may be a good idea to open up a separate issue for that in case I can't get to it right away?

JakeGinnivan:
@mgroves I finally had a chance to look into this

The reason the net35 tests are failing is that net35 doesn't have expression visitors, so we can't format the expression tree nicely. I am skipping these tests for net35.

JakeGinnivan:
And yeah, the docs are in sphinx. It is kinda like markdown just a lot more powerful.

It allows us to pull in external content, so when we update our error messages in shouldly out docs automatically get the changes :) You should be able to follow the patterns we have pretty easily to add docs.

JakeGinnivan:
Thanks for all your work, it will be in the 2.7.0 release (beta 2 will be out within the hour)

PR #266, opened 2015-09-01
Updates NuGet tags for more search scenarios

AmadeusW:
Updates NuGet tags for more search scenarios
Removes a duplicate tag
Fixes a typo

Fixes #265

PR #354, opened 2016-01-13
Updates NuGet tags and fixes a typo

### LINKED ISSUE #265: Nuget package has incorrect tags
The tags contain `shoulda` (a typo?) and duplicate `testing`

screenshot

I should be able to find this package on NuGet by searching for `should`.

AmadeusW:
I was trying to get shouldly via NuGet and I couldn't find it after searching for `should`

It turns out there was a change to the code base that moved the **shouldly.nuspec** file from `src/Shouldly/` to `src/`, but used an old version of the file.

This PR restores changes from #266 (which fixed #265).
Also, you should double check if there were any other changes that might have been lost when transitioning from `src/Shouldly/` to `src/`


JosephWoodward:
Nice catch! Thanks for for the PR.

Issue #308, opened 2015-09-29
Migrate docs to readthedocs.org


JakeGinnivan:
What needs to be done, post here if you are picking up a topic. I will put your name next to the task, please try and submit a pull request soon after so the task does not get blocked. If you cannot complete your task, let me know and I can remove you!

Current docs are at http://docs.shouldly-lib.net/v2.4.0 - some docs are missing, others out of date. Once the section has been created in the new docs, I will delete it out of the current docs site. If you notice I have missed something, yell out!

The docs are in the `docs` folder in the repository, so just submit a pull request. For new files make sure you update the [mkdocs.yml](https://github.com/shouldly/shouldly/blob/master/mkdocs.yml) file. https://github.com/shouldly/shouldly/tree/master/docs

Top level is a single markdown file, second level indicates a H2 (or ##) in that file
## Equality
- [x] ShouldBe
  - [x] Object
  - [x] Numeric
  - [x] DateTime(Offset)
  - [x] TimeSpan
  - [x] String (should link to string ShouldBe section)
  - [x] Enumerable (link to ShouldBe in enumerable section)
- [x] ShouldNotBe
  - [x] Object
  - [x] Numeric
  - [x] DateTime(Offset)
  - [x] TimeSpan
- [x] ShouldBeTrue/False
- [x] ShouldBeNull
  - [x] ShouldNotBeNull
- [x] ShouldBeSameAs
  - [x] ShouldNotBeSameAs
- [x] ShouldBeOneOf
  - [x] ShouldNotBeOneOf
- [x] ShouldBeGreater/LessThan
  - [x] ShouldBeGreaterThan
  - [x] ShouldBeLessThan
  - [x] ShouldBeGreaterThanOrEqualTo
  - [x] ShouldBeLessThanOrEqualTo
- [x] ShouldBeInRange
  - [x] ShouldNotBeInRange
- [x] ShouldBeOfType
  - [x] ShouldNotBeOfType
- [x] ShouldBeAssignableTo
  - [x] ShouldNotBeAssignableTo
## Strings

All these should show the case sensitivity options.
- [x] ShouldBe
  - [x] ShouldNotBe
- [x] ShouldStartWith
  - [x] ShouldNotStartWith
- [x] ShouldEndWith
  - [x] ShouldNotEndWith
- [x] ShouldMatch
  - [x] ShouldNotMatch - Overload missing see #312
- [x] ShouldBeNullOrEmpty
[...]

PR #360, opened 2016-02-09
Added example classes to help with documentation

chaitanyagurrapu:
Hi @JakeGinnivan 
Based on our discussion in #349 here is the updated documentation. While I was at it, I also updated most of the other docs as mentioned in #308.

I haven't done strings yet. And looking at the "Exceptions" section there doesn't seem to be a Func<T> in shouldly, only Func<Task>. And I am not entirely clear on what you mean by "ShouldThrow extension method". I haven't done the "ShouldThrowAsync" either.

I wanted to submit this PR before it got too big for review and to make sure I was still following the intended roadmap regarding documentation. It seems to be generating the right .txt files under "src\DocumentationExamples\CodeExamples", but I haven't got the full documentation tool chain set up so I can't see it all laid out. Also not sure if the "docs\assertions\*.rst" files need to be created for each one or if they are generated as part of the build/deploy chain somewhere.


JosephWoodward:
Awesome work :+1: It looks good from what I can see. Are you able to commit the generated text files too? I'm happy to make a start on adding the .rst docs for them.

chaitanyagurrapu:
Done @JosephWoodward .

Although it seems like one of the shouldly methods (ShouldCompleteIn) seems to be failing. The associated documentation examples test, which is a copy/paste of the actual shouldly unit test is not behaving as expected. Even though the shouldly unit test seems to be working fine. Not sure why. Will investigate later, but if you have any clues that would be great.

JosephWoodward:
Thanks for doing that @chaitanyagurrapu. I had exactly the same issue with the same test failing last night whilst I was sorting out another PR. I re-ran the test it seemed to run without any problems which was odd.

Is the test repeatedly failing for you? I'd be interested in seeing the output you're getting.

chaitanyagurrapu:
Fails pretty consistently for me. I am using the ms test runner that comes with visual studio rather than the resharper or any other third party version. I am guessing that is the test that is failing in the appveyor CI build as well? This is my output.

image

JosephWoodward:
You're right, I'm getting the same test failing. I must be getting confused with another test I had that was temporarily failing. I'll raise this as an issue.

chaitanyagurrapu:
Found the problem and fixed it @JosephWoodward

JosephWoodward:
Oh amazing! Nice one! :+1:

JosephWoodward:
There we go, all merged. Thanks so much for that :)

PR #362, opened 2016-02-18
remove apostrophe

chrissutton:
switch to possessive its


JosephWoodward:
Nice catch! Thanks for this.

PR #368, opened 2016-04-02
Added ShouldHaveSingleItem for Enumerables.

### LINKED ISSUE #365: ShouldHaveSingleItem()
I often have this in my code:

```
var list = ...
list.Count.ShouldBe(1);
var item = list[0]; // or list.Single();
item.Foo.ShouldBe("Bar");
```

I would like a shouldly assertion:

```
var item = list.ShouldHaveSingleItem();
item.Foo.ShouldBe("Bar");
```

The important thing is the ShouldHaveSingleItem method should return the single item.

ssiko:
Resolve #365 - ShouldHaveSingleItem enhancement, IEnumerable extension method.

### DIFF
--- a/src/Shouldly.Shared/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
+++ b/src/Shouldly.Shared/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
@@ -127,7 +127,7 @@ namespace Shouldly
 
         public static void ShouldBeEmpty<T>(this IEnumerable<T> actual, [InstantHandle] Func<string> customMessage)
         {
-            if (actual == null || (actual != null && actual.Count() != 0))
+            if (actual == null || actual.Any())
                 throw new ShouldAssertException(new ExpectedShouldlyMessage(actual, customMessage).ToString());
         }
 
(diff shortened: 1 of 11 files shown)


src/Shouldly.Shared/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
```diff
@@ -147,6 +147,24 @@ public static void ShouldNotBeEmpty<T>(this IEnumerable<T> actual, [InstantHandl
...
+        }
+
+        public static T ShouldHaveSingleItem<T>(this IEnumerable<T> actual, string customMessage)
+        {
+            return ShouldHaveSingleItem(actual, () => customMessage);
+        }
+
+        public static T ShouldHaveSingleItem<T>(this IEnumerable<T> actual, [InstantHandle] Func<string> customMessage)
+        {
+            if (actual == null || (actual != null && actual.Count() != 1))
```
JosephWoodward: Do we need the second null reference check? Wouldn't  `actual == null || actual.Count() != 1` suffice?

ssiko: That should suffice. I did see that same pattern in these Enumerable Extension Methods as well:

```cs
        public static void ShouldNotBeEmpty<T>(this IEnumerable<T> actual, [InstantHandle] Func<string> customMessage)
        {
            if (actual == null || actual != null && !actual.Any())
                throw new ShouldAssertException(new ExpectedShouldlyMessage(actual, customMessage).ToString());
        }
```
And
```cs
        public static void ShouldBeEmpty<T>(this IEnumerable<T> actual, [InstantHandle] Func<string> customMessage)
        {
            if (actual == null || (actual != null && actual.Count() != 0))
                throw new ShouldAssertException(new ExpectedShouldlyMessage(actual, customMessage).ToString());
        }
```

Should I update these as well in this PR? Also seems like we could use ```Any()``` in ```ShouldBeEmpty()``` as well for consistency.

PR #374, opened 2016-05-05
Addition of Should.Throw and Should.ThrowAsync methods

StuartFerguson:
To support passng in the Exception type as a parameter rather than as a type argument.

This enables easier use with NUnit/xUnit theory tests where different values may throw different exception types

### DIFF
--- a/src/Shouldly.Shared/ShouldStaticClasses/ShouldThrow.cs
+++ b/src/Shouldly.Shared/ShouldStaticClasses/ShouldThrow.cs
@@ -41,6 +41,39 @@ namespace Shouldly
             throw new ShouldAssertException(new ShouldlyThrowMessage(typeof(TException), customMessage, shouldlyMethod).ToString());
         }
 
+        /*** Should.Throw(Action) ***/
+        public static Exception Throw([InstantHandle] Action actual, Type exceptionType)
+        {
+            return Throw(actual, () => null, exceptionType);
+        }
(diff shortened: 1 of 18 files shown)


StuartFerguson:
In looking at the Build History this issue seems to have been introduced in the last merge is it possible someone can advise what needs to be done to rectify this issue?

Regards,

Stuart

JosephWoodward:
Yes, I noticed this too. I can't see any reason why the [last merge](https://github.com/shouldly/shouldly/pull/372/files) would have broken anything. I'm hoping to have a look at it tonight to see if I can get it sorted.

StuartFerguson:
@JosephWoodward did you get a chance to look at the build error from this pull request?

JosephWoodward:
Hi @StuartFerguson I've finally gotten around to having a look into this and for the life of me can't figure out why. I don't see the error messages locally so wonder if it's something to do with AppVeyor. I don't suppose you've seen this before @JakeGinnivan ?

JakeGinnivan:
Pretty sure this was due to a bad release of GitVersion unfortunately. Will rebuild this PR and check

JakeGinnivan:
@StuartFerguson this looks good.

The only comment is do you think it's worth switching the generic overloads over to using your new ThrowInternal to reduce code duplication?

JakeGinnivan:
Thanks @StuartFerguson will be out in 2.8.0

PR #380, opened 2016-05-22
Renaming the AssemblyInfo.cs because GitVersion now adds in the missi…

JakeGinnivan:
…ng attributes into all AssemblyInfo.cs files


JosephWoodward:
Oh great, this one had me stumped! How would it have broken the build when no packages were updated?

PR #383, opened 2016-05-30
Trim quotes when initializing DiffTools

alexandernyquist:
This fix trims quotes when initializing DiffTools. It may not be the correct solution (depending on your view) but it fixes the most common problem.

I could strip all characters from `Path.InvalidPathChars` but I think that may cause some wierd behavior, since they are allowed in the environment variables.


JakeGinnivan:
Thanks! :)

PR #384, opened 2016-06-09
RC2 Update - WIP

JosephWoodward:
Initial pull request


JakeGinnivan:
I have opened another PR which moves away from the shared project and PCL approach to using all the new tooling. I can push the branch into the main repo if you want then we can both try and push it forward

PR #386, opened 2016-06-09
Switch to dotnet cli RC2 Preview 1 Tooling

JakeGinnivan:


JakeGinnivan:
@JosephWoodward I think this is pretty much done. I am still going to introduce Cake to do the builds which should fix the build issues.

Let me know if this all works for you or if you have any suggestions

JosephWoodward:
@JakeGinnivan Great! I was hoping to have helped you and sat down last night to have a look at it but wasn't sure where to start.

JakeGinnivan:
I got some help while at NDC, made it easier to get started

khellang:
Nice! That looks like a pretty clean port :smile: I'don't proyally ignore the lockfiles though.

JakeGinnivan:
@khellang yeah, most of the hard work was making it PCL which was 2.7.0 (which worked with RC1 I am pretty sure). Just switched to feature flags and cleaned a few things up.

The problem with not committing lock files is like not doing npm shrinkwrap right? Your transitive dependencies float? Is there a good workflow for updating transitive deps?

khellang:
When you push to NuGet, your dependencies are fixed, I think. So there's no way for transitive dependencies to float. I also think doing a `dotnet restore` might overwrite your lockfiles at any time anyway. There's no way of _actually_ locking them (hint: they're not really a lockfile :stuck_out_tongue:). It's usually just noise when you're checking them in.

JakeGinnivan:
Ok, I think this is ready to go. Am heading to bed but will merge and try and get a (automated) release out tomorrow :)

Any thoughts on where I should put the deployment process so any contributors can push it out?

JosephWoodward:
Awesome work Jake! :+1: :+1:

khellang:
:tada: :clap: :cake: :balloon: :grin: :dancers:

JakeGinnivan:
:shipit:

JakeGinnivan:
bet that bites me

PR #393, opened 2016-07-29
Adding VisitTypeBinary override for ExpressionStringBuilder - Resolves #389

### LINKED ISSUE #389: ShouldAllBe does not show the element predicate correctly for `is` operator
The following snippet produce incorrect assertion message:

			batchCommand.Commands.ShouldAllBe(command=>command is CreateResourceCommand);

```
Shouldly.ShouldAssertException : batchCommand.Commands
    should satisfy the condition
command
    but
[Tp.Rest.Commands.UpdateResourceCommand (22668405), Tp.Rest.Commands.UpdateResourceCommand (29839228)]
    do not
```

I expect a `command is CreateResourceCommand` as an assertion message

ssiko:
Threw together this PR for #389 . Seems like we were yet to override ```VisitTypeBinary()``` in the ```ExpressionVisitor``` class, so it was not returning the full expression.

### DIFF
--- a/src/Shouldly.Tests/ShouldAllBe/TypeBinaryExpressionScenario.cs
+++ b/src/Shouldly.Tests/ShouldAllBe/TypeBinaryExpressionScenario.cs
@@ -1,0 +1,46 @@
+﻿using Shouldly.Tests.Strings;
+using System.Collections.Generic;
+using Xunit;
+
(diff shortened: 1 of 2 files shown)


JosephWoodward:
Awesome, thanks!

Issue #364, opened 2016-02-22
Allow custom IComparer instances for classes that don't implement IComparable


josemesona:
I have a 3rd party class that I can manually compare but it doesn't implement IComparable.  So I want to be able to call `ShouldBeGreaterThan<T>()` but this doesn't seem possible since that method requires that `T` implement `IComparable`.  Is there another way?  I was thinking there could be another parameter to methods like `ShouldBeGreaterThan<T>`() of `type IComparer<T>` and Shouldly would use that class for the comparison instead of calling `IComparable.Compare()`.  Does that seem like something worth implementing?

``` CSharp
public static void ShouldBeGreaterThan<T>(this T actual, T expected, IComparer<T> comparer, Func<string> customMessage)
{
    // Call Shouldy.Is() and which uses custom comparer
}
```

For now I have my own extension methods like this:

``` CSharp
public static void ShouldBeGreaterThan(this ElementLocation actual, ElementLocation expected, Func<string> customMessage)
{
    actual.Line.ShouldBeGreaterThan(expected.Line, customMessage);
}
```

I end up passing the property I want compared but I don't want to have to implement something like this for every class.

Thoughts?

JakeGinnivan:
This sounds like a useful feature

ankurMalhotra:
Given this a go. Please let me know if any improvements can be made :)

eliziario:
This one also could also be closed

JosephWoodward:
Thanks again @eliziario

PR #401, opened 2016-08-31
Compare Custom classes that don't implement IComparable (#364)

ankurMalhotra:
Allow custom IComparer instances for classes that don't implement IComparable.

https://github.com/shouldly/shouldly/issues/364

### DIFF
--- a/src/Shouldly.Tests/ShouldBeLessThanOrEqualTo/CustomObjectScenario.cs
+++ b/src/Shouldly.Tests/ShouldBeLessThanOrEqualTo/CustomObjectScenario.cs
@@ -1,17 +1,17 @@
 ﻿using Shouldly.Tests.TestHelpers;
 using Xunit;
 
-namespace Shouldly.Tests.ShouldBeGreaterThan
+namespace Shouldly.Tests.ShouldBeLessThanOrEqualTo
 {
     public class CustomObjectScenario
     {
         [Fact]
         public void CompareCustom()
         {
-            var customA = new Custom { Val = 2 };
+            var customA = new Custom { Val = 1 };
             var customB = new Custom { Val = 1 };
             var comparer = new CustomComparer<Custom>();
-            customA.ShouldBeGreaterThan(customB, comparer);
+            customA.ShouldBeLessThanOrEqualTo(customB, comparer);
         }
     }
 }
--- a/src/Shouldly.Tests/ShouldBeLessThanOrEqualTo/CustomObjectScenario.cs
+++ b/src/Shouldly.Tests/ShouldBeLessThanOrEqualTo/CustomObjectScenario.cs
@@ -1,4 +1,5 @@
-﻿using Shouldly.Tests.TestHelpers;
+﻿using Shouldly.Tests.Strings;
+using Shouldly.Tests.TestHelpers;
 using Xunit;
 
 namespace Shouldly.Tests.ShouldBeLessThanOrEqualTo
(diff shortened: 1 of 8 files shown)


src/Shouldly/ShouldlyExtensionMethods/ShouldBeGtLtTestExtensions.cs
```diff
@@ -10,6 +11,20 @@ public static partial class ShouldBeTestExtensions
             ShouldBeGreaterThan(actual, expected, () => null);
         }
 
+        public static void ShouldBeGreaterThan<T>(this T actual, T expected, IComparer<T> comparer)
```
josemesona: Do you want add the overloads to LessThan as well?

josemesona:
Looks great to me, just needs LessThan and Equals overloads.

ankurMalhotra:
Cool. Done!

JosephWoodward:
Looks great @ankurMalhotra, thanks for your work! Just one thing though - would you mind updating the tests so they follow our existing tests? [See these tests for example](https://github.com/shouldly/shouldly/blob/master/src/Shouldly.Tests/ShouldBeGreaterThan/EqualScenario.cs).

Thanks!

JakeGinnivan:
I just wanted to expand on @JosephWoodward's comment @ankurMalhotra 

Our tests actually assert the error messages when pdb's are available and when not available. Often by looking at the error messages we can spot how they can be improved.

ankurMalhotra:
Hmm I see. I've the updated the tests. Not completely sure I've done the tests right so please code review.
Thanks

JosephWoodward:
Thanks for that @ankurMalhotra I'll try and get it reviewed this weekend.

src/Shouldly.Tests/ShouldBeLessThanOrEqualTo/CustomObjectScenario.cs
```diff
@@ -0,0 +1,48 @@
...
+            Verify.ShouldFail(() =>
+customA.ShouldBeLessThanOrEqualTo(customB, comparer, "Some additional context"),
+
+errorWithSource:
+@"customA
+    should be less than or equal to
+Shouldly.Tests.TestHelpers.Custom (000000)
+    but was
+Shouldly.Tests.TestHelpers.Custom (000000)
+
```
JosephWoodward: It would be good to output the value of customA and customB, I propose an error message like:

**With source:**
"customA (2) should be less than or equal to customB (1) but was not"

**Without source**
"2 should be less than or equal to 1 but was not"

What are your thoughts @JakeGinnivan ?

JosephWoodward: Hi @ankurMalhotra, that's right - you can see a list of existing message generators [here](https://github.com/shouldly/shouldly/blob/master/src/Shouldly/ShouldlyMessage.cs#L213).

You're right, being a comparer you won't be able to get the value so just drop the value from the final message:

"customA should be less than or equal to customB but was not"

If you need any help with it or have any questions then feel free to let me know :)

JakeGinnivan: Yeah, this is not really possible without either overriding ToString or us trying to print out the values of members which could be too noisy.

JakeGinnivan:
Great work!

PR #404, opened 2016-09-22
Ability to check exception type and exception message.

dcomartin:
Quick extension method to save on having to do two asserts on the exception type and the exception message.  Another possibility is having ShouldThrowAsync<TException>().WithMessage(string) similar to FluentAssertions?


dcomartin:
Fixing the build issues related to extensions and overloads.

JakeGinnivan:
Unsure about this.. it is a simple one liner

```
Should.Throw<FooException>(() => ...).Message.ShouldBe(".....")
```

If you do want this to land we need some tests which will print out the error messages so we can see if this improves the error reporting of ShouldThrow?

Issue #363, opened 2016-02-19
Should.Throw works incorrectly for TimeoutException


steck:
Following test fails

```
[Test]
public void ShouldHandleTimeoutException()
{
    Func<Task> func = async () => { throw new TimeoutException(); };

    Should.Throw<TimeoutException>(func);
}
```

with following message

```
Shouldly.ShouldCompleteInException : 
    Task
        should complete in
    00:00:10
        but did not
  ----> System.TimeoutException : The operation has timed out.
```

AdityaTulasi:
I am getting an exception saying shouldThrow method should throw an exception of type System.Exception but did not throw

bmpj13:
I've been checking out this bug. The problem is that TimeoutException has a special meaning inside Shouldly's.
Should.Throw catches TimeoutException to tell it couldn't run the task in the set time.
An easy fix, I think, would be creating Shouldly's own exception timeouts, so it wouldn't get mixed with the ones available to any user.

Let me know what you think, as I can try to work on it.

eliziario:
looks like this could be closed, as it seems to me that #363 fixes it.

JosephWoodward:
Oh, thanks for point that out @eliziario!

PR #413, opened 2016-11-27
Issue#363

bmpj13:
Code fixing [issue#363](https://github.com/shouldly/shouldly/issues/363). The idea is to have a Shouldly's own TimeoutException, so it doesn't get mixed with the TimeoutException provided by .NET Framework


bmpj13:
Ok, so CI detects a fail in a test... I don't know why, since the test is accepted in my machine.
I'll look into it, but do you have any idea what it might be?

PR #414, opened 2016-11-27
Issue#363

bmpj13:
Tackling [issue#363](https://github.com/shouldly/shouldly/issues/363) . The idea is to create Shouldly's own TimeoutException, so it doesn't get mixed with the TimeoutException provided by .NET Framework.


bmpj13:
Sorry guys, I closed the [other PR](https://github.com/shouldly/shouldly/pull/413) because I accidentally commited a file that wasn't supposed to be there :smile:

src/Shouldly/ShouldlyTimeoutException.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace Shouldly
+{
+    public class ShouldlyTimeoutException : Exception
```
JosephWoodward: Would it be better to extend `TimeoutException` in this instance?

JosephWoodward: What effect (if any) would this have on backwards compatibility? (@JakeGinnivan may be able to answer this as he knows everything)

bmpj13: After some testing, I think it, in fact, would be better to extend TimeoutException. Not only for backwards compatibility, but also to enhance Shouldly's messages.


The following code
```
[Test]
public void ShouldHandleTimeoutException()
{
    var task = new Func<Task>(() => {System.Threading.Thread.Sleep(20000); 
                                     return new Task(null); });
    task.ShouldThrow(typeof(Exception));
}
```

With my code claims that "Shouldly.ShouldlyTimeoutException was thrown". 
When extending TimeoutException it says that the operation exceeded the time limit, which is clearer.

JosephWoodward: Sounds better 👍

bmpj13:
@JosephWoodward I don't know why it is failing. The test *FuncOfTaskOfTWithTimeoutScenario* passes in my machine... Can you take a look at it?

bmpj13:
Ok so apparently it had to do with the .NET version... all good now :+1:

JakeGinnivan:
LGTM, :shipit:

JosephWoodward:
Thanks for this @bmpj13! Merged 👍

Issue #415, opened 2016-12-08
NullReferenceException when using ShouldNotHaveFlag


niklaskallander:
Hi,

I'm getting a NullReferenceException when using the ShouldNotHaveFlag assertion. Nothing is null on the specific code-line where I use it. However, the enum value does in fact have the flag I'm expecting it **not** to have (if that helps).

**Shouldly stacktrace:**
```
at Shouldly.Internals.ActualCodeTextGetter.ParseStackTrace(StackTrace trace)
at Shouldly.Internals.ActualCodeTextGetter.GetCodeText(Object actual, StackTrace stackTrace)
at Shouldly.ShouldlyAssertionContext..ctor(String shouldlyMethod, Object expected, Object actual, StackTrace stackTrace)
at Shouldly.ExpectedActualShouldlyMessage..ctor(Object expected, Object actual, Func`1 customMessage, String shouldlyMethod)
at Shouldly.ShouldlyExtensionMethods.ShouldHaveEnumExtensions.ShouldNotHaveFlag(Enum actual, Enum expectedFlag, Func`1 customMessage)
at Shouldly.ShouldlyExtensionMethods.ShouldHaveEnumExtensions.ShouldNotHaveFlag(Enum actual, Enum expectedFlag)
```

**Code for reproducing the issue:**
```
[Flags]
public enum ReproEnum
{
    Value1 = 1,
    Value2 = 2,
    Value3 = 4,
    All = Value1 | Value2 | Value3
}

[Test]
public void Repro()
{
    var values = ReproEnum.All;

    values.ShouldNotHaveFlag(ReproEnum.Value2);
}
```

Thanks!

JosephWoodward:
Thanks for reporting this, we always welcome pull requests so if you do fancy giving it ago it would be most appreciated.

niklaskallander:
I'll have a look-see, no promises! (:

niklaskallander:
I've created a pull request for this: https://github.com/shouldly/shouldly/pull/416, but this will only fix the issue for my particular case/scenario. If the ShouldyMethods attribute is missing anywhere else those assertions would also fail (if the assertions won't pass). 

However, I do believe fixing ActualCodeTextGetter.ParseStackTrace should be a separate issue/pr?

niklaskallander:
So, obviously, my tiny (and in my eyes, innocent) code-change caused a cascade of test failures. Where did I go wrong? (:

niklaskallander:
[...]

PR #416, opened 2016-12-10
Fix for #415 (ShouldlyMethods attribute were missing on ShouldBeEnumE…

niklaskallander:
…xtensions class, causing ActualCodeTextGetter.ParseStackTrace to fail in finding the code frame of the assertion).


JosephWoodward:
Thanks for the pull request @niklaskallander, sorry it's taken us so long to get it merged.

Issue #282, opened 2015-09-28
Implement ShouldEndWith() for IEnumerable


ceregador:
_ShouldEndWith(object element)_
Asserts that the collection ends with the specified element.

JakeGinnivan:
Linq is pretty powerful, I don't think we should try and replace it.

`enumerable.LastOrDefault().ShouldBe(element)` will be just as good.

SurajGupta:
ShouldEndWith is far more readable than the current solution.  Same comment as here: https://github.com/shouldly/shouldly/issues/280

ceregador:
`var collection = new[] {1, 3, 5};`
`collection.ShouldEndWith(3);`

_Example of an error message for ShouldEndWith():_
Expected collection to end with 3, but found 5.

Issue #281, opened 2015-09-28
Implement ShouldStartWith() for IEnumerable


ceregador:
_ShouldStartWith(object element)_
Asserts that the collection starts with the specified element.

JakeGinnivan:
Can be achieved via linq. `enumerable.FirstOrDefault().ShouldBe(element)`

SurajGupta:
ShouldStartWith is far more readable than the current solution.  Same comment as here: https://github.com/shouldly/shouldly/issues/280

ceregador:
`var collection = new[] {1, 3, 5};`
`collection.ShouldStartWith(3);`

_Example of an error message for ShouldStartWith():_
Expected collection to start with 3, but found 1.

bmpj13:
I can take a look at this, if you want @JakeGinnivan

brunomasantos:
Hi guys,

I wanted to start contributing to this project. Wondering if is this that i can try to pick up. I see its open a long time now.

Thanks guys

JosephWoodward:
@brunomasantos We already have an open [pull request](https://github.com/shouldly/shouldly/pull/417) surround this one.

brunomasantos:
Completely missed it, sorry 👍

JosephWoodward:
I'm going to close this one off as we've not had much interest it as a feature and agree with Jake that `enumerable.FirstOrDefault().ShouldBe(element)` is clearer.

PR #417, opened 2016-12-14
ShouldBeginWith & ShouldEndWith

bmpj13:
Here's my solution to implement the predicates proposed on #281 and #282. It uses Linq to access the first/last element of the enumerable.

### DIFF
--- a/src/Shouldly/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
+++ b/src/Shouldly/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
@@ -43,22 +43,20 @@ namespace Shouldly
                 throw new ShouldAssertException(new ExpectedActualShouldlyMessage(expected, actual, customMessage).ToString());
         }
 
-        public static void ShouldStartWith<T>(this IEnumerable<T> actual, T expected)
+        public static void ShouldBeginWith<T>(this IEnumerable<T> actual, T expected)
         {
-            ShouldStartWith(actual, expected, () => null);
+            ShouldBeginWith(actual, expected, () => null);
         }
 
-        public static void ShouldStartWith<T>(this IEnumerable<T> actual, T expected, string customMessage)
+        public static void ShouldBeginWith<T>(this IEnumerable<T> actual, T expected, string customMessage)
         {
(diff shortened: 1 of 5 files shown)


bmpj13:
Hey guys, here's my solution to ShouldBegin/EndWith. I haven't implemented it with tolerance yet, because I wanted to know what do you think of the solution right now.

Btw, I'm using Linq's First and Last, instead of the First/LastOrDefault, because if it goes to default, then it can lead do some mistakes. For example:
```
[Fact]
public void BeginShouldPass()
{
      int[] target = {};
      target.ShouldBeginWith(0);
}
```
Passes because FirstOrDefault defaults to 0. Using only First, it will throw InvalidOperationException	, stating that the enumerable is empty... I wanted to point it out to know what you guys think :smile:

JosephWoodward:
Sorry it's taken me so long to get around to this, it's been a busy few months! Out of interest, what was your reasoning to go with `ShouldBeginWith` vs `ShouldStartWith`?

src/Shouldly/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
```diff
@@ -43,6 +43,38 @@ public static void ShouldNotContain<T>(this IEnumerable<T> actual, T expected, [
...
+        public static void ShouldBeginWith<T>(this IEnumerable<T> actual, T expected, string customMessage)
+        {
+            ShouldBeginWith(actual, expected, () => customMessage);
+        }
+
+        public static void ShouldBeginWith<T>(this IEnumerable<T> actual, T expected, [InstantHandle] Func<string> customMessage)
+        {
+            if (!actual.First().Equals(expected))
+                throw new ShouldAssertException(new ExpectedActualShouldlyMessage(expected, actual, customMessage).ToString());
+        }
```
hrai: .NET API uses `StartsWith` so perhaps we should follow the heuristics and rename the methods to `ShouldStartWith`

JosephWoodward:
I'm going to close this one off as we've not had much interest it as a feature and agree with Jake that `enumerable.FirstOrDefault().ShouldBe(element);` is clearer. Thanks for putting the pull request together.

PR #423, opened 2017-02-20
Fix for #422 (Try catch around environment variable calls)

SteelFlow:
I understand that my PATH environment variables shouldn't include "illegal characters". With that said, i here put a Try/catch around the integration with local computer.

### DIFF
--- a/src/Shouldly/Configuration/DiffTool.cs
+++ b/src/Shouldly/Configuration/DiffTool.cs
@@ -52,26 +52,33 @@ namespace Shouldly.Configuration
 
         static string GetFullPath(string fileName)
         {
+            if (File.Exists(fileName))
+                return Path.GetFullPath(fileName);
+
+            var processPath = Environment.GetEnvironmentVariable("PATH", EnvironmentVariableTarget.Process);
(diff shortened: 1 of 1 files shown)


slang25:
How about adding a class like this:
```csharp
using System;
using System.Collections.Generic;

namespace Shouldly
{
    internal static class EnumerableExtensions
    {
        public static IEnumerable<T> CatchExceptions<T> (this IEnumerable<T> src)
        {
            using (var enumerator = src.GetEnumerator())
            {
                bool next = true;

                while (next) {
                    try {
                        next = enumerator.MoveNext();
                    } catch (Exception) {
                        // Ignore exception
                        continue;
                    }

                    if (next) {
                        yield return enumerator.Current;
                    }
                }
            }
        }
    }
}
```

Then you can do this:
```csharp
return values.Split(';')
                .Where(p => !string.IsNullOrEmpty(p))
                .Select(path => path.Trim('"'))
                .Select(path => Path.Combine(path, fileName))
                .CatchExceptions()
                .FirstOrDefault(File.Exists);
```
Which will be more forgiving.

SteelFlow:
How about this? A TryCombine method that will ignore the illegal combines.

JosephWoodward:
Thanks for the PR @SteelFlow!

PR #432, opened 2017-04-10
Tooling migration

JosephWoodward:
Initial PR for project.json > .csproj migration


slang25:
@JosephWoodward  `<DebugType>full</DebugType>` Should fix most failing tests

JosephWoodward:
Hmm, still really stuck with this error, anyone have anyone seen it before?
screen shot 2017-06-29 at 11 38 20

cc @JakeGinnivan

jeffkl:
@JosephWoodward I work on the MSBuild team at Microsoft.  Now that you're switching to the MSBuild-based projects and away from `xproj`, the Cake script probably doesn't need to run [DotNetCorePack](https://github.com/shouldly/shouldly/blob/master/build.cake#L61).  You can have a NuGet package created at build time by setting an MSBuild property `GeneratePackageOnBuild` to `true`.  Here's an example:  http://blog.myget.org/post/2017/03/15/visual-studio-2017-and-net-core-support-on-myget.aspx

If the package is created during build, you won't need to run pack as a separate phase but if you do want to, you would now run `msbuild /Target:Pack`.  I'm not very familiar with Cake but I'm sure there's a way to do that.

JosephWoodward:
Thanks a bunch @jeffkl, I'd missed that part! I'm still getting the SatelliteDllsProjectOutputGroupWithTargetFramework error so I'll keep digging.

```bash
"C:\projects\shouldly\src\Shouldly.sln" (Build target) (1) ->
"C:\projects\shouldly\src\Shouldly\Shouldly.csproj" (default target) (2) ->
"C:\projects\shouldly\src\Shouldly\Shouldly.csproj" (SatelliteDllsProjectOutputGroupWithTargetFramework target) (2:5) ->
  C:\projects\shouldly\src\Shouldly\Shouldly.csproj : error MSB4057: The target "SatelliteDllsProjectOutputGroupWithTargetFramework" does not exist in the project.
```

jeffkl:
Alright I checked out your branch and figured out what's wrong.  You're using `MSBuild.Sdk.Extras` which defines this target `SatelliteDllsProjectOutputGroupWithTargetFramework`.  At restore time, NuGet creates `obj\Shouldly.csproj.nuget.g.targets` that gets imported and it contains the following:

```xml
<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <PropertyGroup>
    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
  </PropertyGroup>
  <ImportGroup Condition=" '$(TargetFramework)' == '' AND '$(ExcludeRestorePackageImports)' != 'true' ">
    <Import Project="$(NuGetPackageRoot)msbuild.sdk.extras\1.0.0\buildMultiTargeting\MSBuild.Sdk.Extras.targets" Condition="Exists('$(NuGetPackageRoot)msbuild.sdk.extras\1.0.0\buildMultiTargeting\MSBuild.Sdk.Extras.targets')" />
  </ImportGroup>
  <ImportGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' AND '$(ExcludeRestorePackageImports)' != 'true' ">
    <Import Project="$(NuGetPackageRoot)msbuild.sdk.extras\1.0.0\build\netstandard1.0\MSBuild.Sdk.Extras.targets" Condition="Exists('$(NuGetPackageRoot)msbuild.sdk.extras\1.0.0\build\netstandard1.0\MSBuild.Sdk.Extras.targets')" />
  </ImportGroup>
</Project>
```

The initial build of a project with `TargetFrameworks` set launches N number of builds with `TargetFramework` set.  For the first build, it imports `buildMultiTargeting\MSBuild.Sdk.Extras.targets` which tells MSBuild to [build a target](https://github.com/onovotny/MSBuildSdkExtras/blob/master/src/buildMultiTargeting/MSBuild.Sdk.Extras.targets#L10).  However, when building the project targeting a single framework, `net35` and `net451` aren't importing the [file that has the necessary target](https://github.com/onovotny/MSBuildSdkExtras/blob/master/src/build/netstandard1.0/NetCoreSdk1Workarounds.targets#L20).

The package supports `netstandard1.0` and `portable-net4+wp7+sl4+win8+wpa81` which I believe should cover all frameworks.  But NuGet doesn't appear to map `portable-net4+wp7+sl4+win8+wpa81` to `net35` or `net461`.  @onovotny might know why that's happening or if we just need to add a `net35` folder to the package with the appropriate imports.

As a workaround, you can define an empty target in `Shouldly.csproj` which is overridden when targeting `netstandard1.0`.  I'm not sure if that breaks things but I don't believe you're using satellite assemblies anyway.

```xml
  <Target Name="SatelliteDllsProjectOutputGroupWithTargetFramework" />

  <Import Project="$(MSBuildSDKExtrasTargets)" Condition="Exists('$(MSBuildSDKExtrasTargets)')" />
```

global.json
```diff
@@ -1,4 +1,3 @@
 {
-    "projects": [ "src" ],
-    "sdk": { "version": "1.0.0-preview2-003131" }
+    "sdk": { "version": "1.0.0" }
```
onovotny: I would not use 1.0.0. Use 1.0.4 at the very least. Many bugs fixed.

NuGet.config
```diff
@@ -5,6 +5,7 @@
     <!-- add key="CoreCLR Xunit" value="https://www.myget.org/F/coreclr-xunit/api/v3/index.json" / -->
     <!-- add key="AspNet CI Feed" value="https://www.myget.org/F/aspnetcirelease/api/v3/index.json" /> -->
     <add key="myget.org xunit" value="https://www.myget.org/F/xunit/api/v3/index.json" />
+    <add key="myget.org msbuilsdkdextras" value="https://www.myget.org/F/msbuildsdkextras/api/v3/index.json" />
```
onovotny: No need for this. NuGet.org has the latest version.

src/Shouldly.Tests/Shouldly.Tests.csproj
```diff
@@ -0,0 +1,37 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net452</TargetFramework>
+    <AssemblyName>Shouldly.Tests</AssemblyName>
+    <PackageId>Shouldly.Tests</PackageId>
+    <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
```
onovotny: I don't think this is required

src/Shouldly.Tests/Shouldly.Tests.csproj
```diff
@@ -0,0 +1,37 @@
...
+  <ItemGroup>
+    <Compile Remove="**\*.approved.cs;**\*.received.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Shouldly\Shouldly.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0-preview-20170106-08" />
```
onovotny: I'd use either the stable version or the latest 15.3 preview test sdk version.

src/Shouldly/Shouldly.csproj
```diff
@@ -0,0 +1,72 @@
...
+  <PropertyGroup>
+    <Description>Shouldly - Assertion framework for .NET. The way asserting *Should* be</Description>
+    <TargetFrameworks>net35;net451;netstandard1.3</TargetFrameworks>
+    <AssemblyName>Shouldly</AssemblyName>
+    <PackageId>Shouldly</PackageId>
+    <PackageTags>test;unit;testing;TDD;AAA;should;testunit;rspec;assert;assertion;framework</PackageTags>
+    <PackageIconUrl>https://raw.github.com/shouldly/shouldly/master/Icons/package_icon.png</PackageIconUrl>
+    <PackageProjectUrl>http://shouldly.github.com</PackageProjectUrl>
+    <PackageLicenseUrl>https://github.com/shouldly/shouldly/blob/master/LICENSE.txt</PackageLicenseUrl>
+    <NetStandardImplicitPackageVersion Condition=" '$(TargetFramework)' == 'netstandard1.3' ">1.6.0</NetStandardImplicitPackageVersion>
```
onovotny: Don't do this -- Remove this line and enable the default implicit ref or 1.6.1 for the netstandard 1.3 tfm. 1.6.0 contains many bugs, esp with Xamarin.

src/Shouldly/Shouldly.csproj
```diff
@@ -0,0 +1,72 @@
...
+    
+    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
+    <IncludeSymbols>True</IncludeSymbols>
+    <IncludeSource>True</IncludeSource>
+
+    <ComVisible>false</ComVisible>
+    <IncludeProjectPriFile>false</IncludeProjectPriFile>
```
onovotny: Not sure why you need this? There shouldn't be one generated anyway

src/Shouldly/Shouldly.csproj
```diff
@@ -0,0 +1,72 @@
...
+  
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'net451' ">
+    <DefineConstants>$(DefineConstants);ShouldMatchApproved;StackTrace;Dynamic;Serializable;Async;ExpressionTrees</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
+    <DefineConstants>$(DefineConstants);Async;Dynamic;NewReflection;ExpressionTrees</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
```
onovotny: Don't do this....this is considered bad practice now. @terrajobst can explain in detail if you want. Remove this item group and leave the default netstandard.library 1.6.1 if the tfm is a netstandard one.

src/Shouldly/Shouldly.csproj
```diff
@@ -0,0 +1,72 @@
...
+    <PackageReference Include="System.Diagnostics.StackTrace" Version="4.3.0" />
+    <PackageReference Include="System.Dynamic.Runtime" Version="4.3.0" />
+    <PackageReference Include="System.IO.FileSystem" Version="4.3.0" />
+    <PackageReference Include="System.Linq" Version="4.3.0" />
+    <PackageReference Include="System.Linq.Expressions" Version="4.3.0" />
+    <PackageReference Include="System.Reflection" Version="4.3.0" />
+    <PackageReference Include="System.Reflection.TypeExtensions" Version="4.3.0" />
+    <PackageReference Include="System.Threading.Timer" Version="4.3.0" />
+    <PackageReference Include="Microsoft.CSharp" Version="4.3.0" />
+    <PackageReference Include="MSBuild.Sdk.Extras" Version="1.0.0" PrivateAssets="all" />
```
onovotny: Not sure why you need the sdk extras at all? It's really only needed to build project types other than net/netstandard/netcoreapp (like Xamarin, UWP, etc).

src/Shouldly/Shouldly.csproj
```diff
@@ -0,0 +1,72 @@
...
+  </ItemGroup>
+
+  <Import Project="$(MSBuildSDKExtrasTargets)" Condition="Exists('$(MSBuildSDKExtrasTargets)')" />
```
onovotny: Can be removed, the extras package shouldn't be needed here.

onovotny: From what I saw, this is a bug you hit in 15.2, should be fine in 15.2.2.

onovotny:
Also, with your `net35` target, you may not be able to use `dotnet build`. You might need `msbuild /t:build`, `msbuild /t:pack`, etc.

JosephWoodward:
@jeffkl @onovotny You guys are amazing, thanks so much for your help! I ran the migration tool so wasn't sure what I needed vs didn't. I'll give your suggestions a try and see how I get on.

jeffkl:
Glad to help. I'll keep an eye on this PR. 

FYI https://github.com/Microsoft/msbuild/pull/2250

src/Shouldly/Shouldly.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>Shouldly - Assertion framework for .NET. The way asserting *Should* be</Description>
-    <TargetFrameworks>net451;netstandard1.3</TargetFrameworks>
+    <TargetFrameworks>net451;net452;netstandard1.3</TargetFrameworks>
```
onovotny: Is there a particular reason you need to support net451 and net452? there's virtually no API surface area different between them; in general, I suggest targeting the fewest frameworks you need to support.

slang25: Agreed, I presume this was temporary in an effort to troubleshoot the ci build.

@JosephWoodward I've PRed a working build. If you are happy to merge it would be good to get some more eyes on this.

JosephWoodward: Hi @onovotny 
It was a desperate attempt to get the build working before popping out, given the tests were running against 452. Thanks for the PR @slang25, that's all green now - good job! 👍

PR #439, opened 2017-07-01
Make diff tools configuration lazy

### LINKED ISSUE #437: EntryPointNotFoundException on rider / .net 462 / macOS
Quick one which should be easy to repro. Using 2.8.3, rider, xunit, .net 4.6.2 classlib, running tests in the runner throws with an EntryPointNotFoundException inside a TypeInitializerException that leads me to
```
      [DllImport("kernel32.dll", SetLastError = true)]
      private static extern IntPtr CreateToolhelp32Snapshot(uint dwFlags, uint th32ProcessId);
```
Which I think means it's trying to do something funky with diff windows in VS or something.

Downgrading to 2.8.2 solves the issue.

slang25:
Fixes #437 

This change makes the VS diff tool not throw when it does the windows specific native interop to find the current running vs process, but more importantly it makes the initialization of all diff tools lazy, so you only pay the cost of doing the work if you actually use the `ShouldMatchApproved` extension method.

It could do with a refactor, but I didn't want to make a breaking change.


JosephWoodward:
Nice,  thanks for this Stu.

slang25:
I removed the lazy polyfill this morning, did you redo it and push over the top?

JosephWoodward:
Really? It was still there in the PR when I pushed it earlier?

slang25:
I'm guessing I didn't push it, I think I'm going crazy.

JosephWoodward:
@jeffkl No problem, was going to do some work on Shouldly tonight so I could get it out in a new beta version.

JosephWoodward:
@jeffkl [3.0.0-beta0003](https://www.nuget.org/packages/Shouldly/3.0.0-beta0003) contains the fix for this issue 👍

Issue #449, opened 2017-07-13
Shouldly.dll not signed (netstandard1.3)


AndyGerlicher:
Using the latest NuGet package (2.8.3):
```
>sn -vf lib\net451\Shouldly.dll

Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Assembly 'lib\net451\Shouldly.dll' is valid

>sn -vf lib\netstandard1.3\Shouldly.dll

Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

lib\netstandard1.3\Shouldly.dll does not represent a strongly named assembly
```

Looks like signing was added a long time ago (#40), but the netstandard1.3 dll is not signed.

We're about to adopt Shouldly for MSBuild (Microsoft/msbuild#2250) and I'd rather not have to suppress that compile warning, especially if it's by mistake. Is this by design or something you're aware of?

JosephWoodward:
Hi @AndyGerlicher, thanks for making us aware of this issue, @slang25 has just contributed a fix for it.

jeffkl:
@JosephWoodward can we please get a release with the signed assembly?  I'd like to merge my PR https://github.com/Microsoft/msbuild/pull/2250

JosephWoodward:
@jeffkl Will do, should get it out today 👍

JosephWoodward:
@jeffkl A pre-release of Shouldly 3.0 is now available on NuGet and can be downloaded [here], the reason it's a 3.0 and a major version bump is because we removed support for .NET 3.5; other than that it's stable.

JosephWoodward:
I'll close this issue, if there are any problems then do feel free to post them again and I'd be more than happy to reopen it :)

AndyGerlicher:
Thanks!! We confirmed it's signed and merged our PR :)

JosephWoodward:
@AndyGerlicher Great to hear! :)

PR #450, opened 2017-07-14
More build improvments

slang25:
Following the latest csproj refactoring, I've added some minor improvements, including assembly signing which addresses #449


JosephWoodward:
Thanks for the PR and fixing the signing issue

PR #462, opened 2017-09-21
Correct VSTS/TFS environment variable.

GianLorenzetto:
VSTS build environment variable is incorrect. More details here https://github.com/Microsoft/vsts-agent/issues/936


JosephWoodward:
Thank you for this, can't see why the tests were failing, I'll take a look.

GianLorenzetto:
Cheers. I did check if there were any tests that would be effected by the change, but I couldn't see anything.

PR #468, opened 2017-10-14
Fix typos in SourceCodeTextGetter.cs

TAGC:
Replace all instances of "parantheses" with "parentheses"


JosephWoodward:
Thanks for this!

PR #478, opened 2018-01-22
Replace Thread.Sleep with Task.Wait

JosephWoodward:


slang25:
Out of curiosity, what was the goal here?

JosephWoodward:
Migrating the tests to .NET Standard bit by bit.

PR #479, opened 2018-01-24
.NET Standard tests

JosephWoodward:


global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "2.0.0"
```
slang25: I reckon this raises the barrier to entry and should be left to your build script

src/Shouldly.Tests/ConventionTests/ShouldlyConventions.cs
```diff
@@ -1,3 +1,4 @@
+#if FullFramework
```
slang25: I did some digging, it looks like the problem we are facing is this:
https://github.com/dotnet/corefx/issues/21079
and is fixed in 2.1:
https://github.com/dotnet/coreclr/pull/14696

It's also probably cleaner to keep the compiler constants to the functionality they toggle instead of being 1-1 with the tfm, cause it would be easy for someone to use this for something other than removing code that relies on symbols working for example.

slang25:
A couple of thoughts added @JosephWoodward

global.json
JosephWoodward: I agree, but as I just discovered a moment ago appveyor fails to run the tests without it, not sure how to get around that.

src/Shouldly.Tests/ConventionTests/ShouldlyConventions.cs
JosephWoodward: Added the stacktrace flag to .NET Standard 2.0 and all the tests passed

JosephWoodward: > It's also probably cleaner to keep the compiler constants to the functionality they toggle instead of being 1-1 with the tfm, cause it would be easy for someone to use this for something other than removing code that relies on symbols working for example

Agreed. Now I've got the tests working I can refine these flags.

JosephWoodward:
See replies @slang25

src/Shouldly.Tests/ConventionTests/ShouldlyConventions.cs
slang25: Nice. What's weird for me is that I pull the branch into a fresh folder (proper fresh), do a restore, build, and test, and it cannot get the filenames (which are required by the tests to load the source), and I can see that the symbols are loaded.

I'm 100% hitting bug dotnet/corefx#21079, but clearly it's not an issue in AppVeyor
😕

JosephWoodward:
@slang25 Interested to pull this down on my Windows machine and see if it all builds/tests pass.

JosephWoodward:
lol

src/Shouldly.Tests/ConventionTests/ShouldlyConventions.cs
EddieGarmon: I was hitting the same, but as per the recommendation, if you copy the *.DiaSymReader.*.dll's from C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\Roslyn to the tests bin directory, this temporarily solves the line number issues, and all the netcoreapp2.0 tests pass.

slang25: @EddieGarmon oh nice, I'll have to give that a try. Thanks for the info 👍

PR #488, opened 2018-02-27
Fix broken .netcoreapp2.0 tests locally

lxalln:
.netcore2.0 pdbs on Windows do not include filename or line number.

This is fixed and will be released in 2.1. However, for the time being
if we switch to using portable pdbs, these then work (just like they do
on linux/mac).

Once we switch the target up to 2.1, this change can be removed.

### DIFF
--- a/src/Shouldly.Tests/Shouldly.Tests.csproj
+++ b/src/Shouldly.Tests/Shouldly.Tests.csproj
@@ -9,6 +9,9 @@
     <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
     <DebugType>full</DebugType>
   </PropertyGroup>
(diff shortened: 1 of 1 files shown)


slang25:
Can we not use portable pdbs for net452 assemblies out of interest?

lxalln:
Using portable pdbs works for net452 as well. I was just trying to make as small a change as I could to get the tests passing locally.

SimonCropp:
@lxalln given netcore 2.0 is now "end of support" https://dotnet.microsoft.com/platform/support/policy/dotnet-core

do you still need this?

JosephWoodward:
@SimonCropp No, this can be closed, thanks.

PR #492, opened 2018-03-17
Fixed Test.csproj Now will work on linux

loneosama:
I tried to run the tests on my linux environment but they didnt work so i noticed that it is trying to target 452 which is not present on linux so i simply put a condition there to help it not interefere. 
Note: 
I was using dotnet core 2.1.101
and and was on ubuntu 16.04


JosephWoodward:
Oh nice, thanks for this.

PR #494, opened 2018-03-19
Now will work on vs 2017 Build 15.6 

loneosama:
vs made a little change, but honestly it is some kind of a bug in vs 2017 that doesn't let it start. Do tell if the problem persists in vs 2017 v15.6


JosephWoodward:
Awesome, thanks for this.

PR #510, opened 2018-07-11
Add WinMerge to known diff tools

bdukes:
`/u` keeps WinMerge from adding the files to the MRU list
`/wl` makes the left file (received) read-only


JosephWoodward:
Nice, thanks for this!

PR #511, opened 2018-07-12
Update solution item paths to remove warning on VS load

dlumpp:
Corrected paths to remove project load warning with message "Some of the properties associated with the solution could not be read."
Noticed a new file that looks like it goes here so added ReleaseNotes.md item too.


JosephWoodward:
Missed this PR! Thanks for this.

PR #516, opened 2018-07-30
Add new Do Not Launch Strategy

bdukes:
This should stop diff tool launching when using
Visual Studio Live Unit Testing.  This check is based on the FAQ from
https://blogs.msdn.microsoft.com/visualstudio/2017/03/09/live-unit-testing-in-visual-studio-2017-enterprise/


bdukes:
No problem, thanks!

PR #527, opened 2018-09-21
Add RepositoryUrl MSBuild property

martincostello:
Add the `RepositoryUrl` and `RepositoryType` MSBuild properties so that the next release of Shouldly will have a link to the repo's source code from `nuget.org`.


JosephWoodward:
Good catch, thanks @martincostello

PR #528, opened 2018-09-24
Build all on mac

slang25:
If you were ok with dropping `net40`, here's how you could build for all targets on macOS, "works on my mac with Rider"™


slang25:
screen shot 2018-09-24 at 14 46 37

JosephWoodward:
> If you were ok with dropping net40

Was thinking of this very thing myself 🤔

slang25:
That said, the other PR works well too. The advantage of this one is that you don't even need mono

JosephWoodward:
Just about to release a beta of 4.0, so seriously considering this.

PR #529, opened 2018-09-24
Build all platforms on macOS, requires mono

slang25:
An alternative to #528, but requires mono is installed


JosephWoodward:
This can now be closed.

PR #530, opened 2018-10-07
Add Should.Throw() overloads for Func

martincostello:
In the process of working on justeat/JustEat.StatsD#126, Shouldly caused some code analysis warnings for unused objects in statements such as the below because the instatiated object is not used:

```csharp
Should.Throw<ArgumentOutOfRangeException>(
    () => new SocketTransport(LocalStatsEndpoint(), (SocketProtocol)42));
```

Whilst this is intentional because the constructor is expected to throw, the Roslyn analyzer does not know this, so raises a warning as it is being used in an `Action` delegate.

Rather than litter the code with suppressions (or turn off the rule entirely as this would then not find real violations), I switched to using xunit's `Assert.Throws(Func<object>)` instead, which does not raise such a warning as as far as the analyzer is concerned the constructed instance is returned to the caller which then uses it to do something.

```csharp
Assert.Throws<ArgumentOutOfRangeException>(
    () => new SocketTransport(LocalStatsEndpoint(), (SocketProtocol)42));
```

To remove the need for such suppression gymnastics or inconsistent assertion method usage, this PR adds various overloads that match the `Action` methods for `Should.Throw()` with variants that take `Func<object>`. The use of `Func<object>` removes the need for the type parameter to be explicitly specified so the return type of the delegate can be anything.

As special-case `Func<Task>` overloads already exist, I _think_ this should be a source and binary compatible change as the overload resolution will remain the same as it is today.

I didn't fully follow the documentation tests, so I may have accidentally done the opposite of what was required in them, but I did run the `copy` commands suggested by the failing assertions after the initial changes and commit them.

I also deleted a Visual Studio `UpgradeLog.htm` that made its way into the repository and updated the `.gitignore` so one doesn't get added by accident again in future,


jnm2:
Fwiw, this is what I've been doing:

```cs
Should.Throw<ArgumentOutOfRangeException>(
    () => _ = new SocketTransport(LocalStatsEndpoint(), (SocketProtocol)42));
```

JosephWoodward:
Nice, thanks for the PR @martincostello 👍

Issue #535, opened 2018-10-23
Ship with SourceLink


jnm2:
Would you consider providing PDBs with SourceLink information? This will make it easy for anyone to drop into Shouldly code any time they want to understand what's going on, and users may return the courtesy by taking some initiative in diagnosing bugs they observe. Creating a repro is much easier when you aren't dealing with a black box!

It's easier than it used to be because the csproj SDK now supports it. The recommended route is https://github.com/dotnet/sourcelink#sourcelink-preview.

(Just ran into an NRE inside Shouldly's ExpressionStringBuilder and wanted to investigate.)

jnm2:
I believe this has been closed by #536. Thanks @slang25, that was fast! 🎉

JosephWoodward:
Yup! I'll get a new release out soon!

jnm2:
@JosephWoodward Cool! Hey, if I fixed https://github.com/shouldly/shouldly/issues/499 really fast, would it make it?

jnm2:
Just in case, no pressure: https://github.com/shouldly/shouldly/pull/538

PR #536, opened 2018-10-23
Sourcelink

slang25:
Addresses #535


JosephWoodward:
Looks like this is what's breaking the build, I presume because the `global.json` file has been removed:

This is the build error I'm seeing: `The specified framework version '2.0' could not be parsed`

PR #542, opened 2018-11-04
Fix build, update cake, and general tidy up

slang25:
I'll try to get the build working here.

I'd like cake up to date, sourcelink enabled, no pinned sdk version, and a green check.


slang25:
Bonus, we now build and run tests on AppVeyor linux images 🎉

slang25:
The only thing I'm not happy about currently is I've enabled `ShouldMatchApproved` for `netstandard2.0`, that code should probably check the platform first.

JosephWoodward:
Nice one, thanks for this! 👍

PR #543, opened 2018-11-04
Only launch diff tools on Windows for now

slang25:


slang25:
I need to fix the approval tests cause this changes the public api surface. I'll have a look tomorrow.

PR #545, opened 2018-11-06
Approve public API

JosephWoodward:


slang25:
Amazing 😃 Thanks for finishing. This makes updating TestStack.ConventionTests easier ironically, cause there the test projects use this, now we can port the tests to netstandard2.0 with no code changes.

PR #548, opened 2018-11-20
[WIP] Use portable pdbs on .NET Framework

slang25:
I suspect this is why on some version on .NET Framework you have to use full pdbs, given they are dead and problematic, I'm hoping this will remove the need for them.


JosephWoodward:
Nice 👍

Issue #554, opened 2019-01-21
Should.ThrowAsync error messages should match Should.Throw messages


demoncodemonkey:
The error messages for Should.ThrowAsync don't seem as detailed as the error messages for Should.Throw. To demonstrate what I mean, try adding these to a file in your project:

```
    public class Package
    {
        public static Task<int> DoSomething()
        {
            throw new DivideByZeroException();
        }
    }

    public class Tests
    {
        [Test]
        public void Test1()
        {
            Should.Throw<TimeoutException>(() => Package.DoSomething());
        }

        [Test]
        public void Test2()
        {
            Should.Throw<TimeoutException>(async () => await Package.DoSomething());
        }

        [Test]
        public async Task Test3()
        {
            await Should.ThrowAsync<TimeoutException>(() => Package.DoSomething());
        }

        [Test]
        public async Task Test4()
        {
            await Should.ThrowAsync<TimeoutException>(async () => await Package.DoSomething());
        }
    }

//requires these packages: NUnit 3.11, NUnit3TestAdapter 3.12, Shouldly 3.02
```

Test1 output: 😃 
> Message: Shouldly.ShouldAssertException : Task \`Package.DoSomething()\`
>     should throw
> System.TimeoutException
>     but threw
> System.DivideByZeroException

Test2 output: 😃 
> Message: Shouldly.ShouldAssertException : Task \`async () => await Package.DoSomething()\`
>     should throw
> System.TimeoutException
>     but threw
> System.DivideByZeroException

Test3 output: 😞 
> Message: System.DivideByZeroException : Attempted to divide by zero.

Test4 output: 😞 
> Message: Shouldly.ShouldAssertException : Shouldly uses your source code to generate its great error messages, build your test project with full debug information to get better error messages
> The provided expression
>     handle aggregate exception
> System.TimeoutException
>     but was
> System.DivideByZeroException

JosephWoodward:
Thanks for raising this, I'll take a look.

gmilani:
Hello @JosephWoodward 
I want to start contributing to Shouldly project. Can I start with this issue?

JosephWoodward:
Hi @gmilani,

Yes! That would be great! Let me know if you have any issues or questions.

hrai:
Should this be closed?

JosephWoodward:
Yes it should, thanks!

wuanunet:
Hi,

It seems there’s a minor mistake in the documentation or example. The correct usage should be:

[...]

PR #555, opened 2019-02-01
Fix in the Should.ThrowAsync error messages and exception handling

gmilani:
#554

### DIFF
--- a/src/Shouldly/ShouldStaticClasses/ShouldThrowTaskAsync.cs
+++ b/src/Shouldly/ShouldStaticClasses/ShouldThrowTaskAsync.cs
@@ -48,39 +48,67 @@ namespace Shouldly
         {
 #if StackTrace
             var stackTrace = new StackTrace(true);
-            return actual().ContinueWith(t =>
+            try
             {
-                if (t.IsFaulted)
-                {
-                    if (t.Exception == null)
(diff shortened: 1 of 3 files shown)


src/Shouldly/ShouldStaticClasses/ShouldThrowTaskAsync.cs
```diff
@@ -48,39 +48,67 @@ public static Task<Exception> ThrowAsync(Task task, [InstantHandle] Func<string>
...
+                        throw new ShouldAssertException(new TaskShouldlyThrowMessage(typeof(TException), customMessage).ToString()
+                            , new TaskCanceledException("Task is cancelled"));
+
+                    throw new ShouldAssertException(new TaskShouldlyThrowMessage(typeof(TException), customMessage).ToString());
+                });
+            }
+            catch (Exception e)
+            {
+                if (e is TException)
+                    return Task.Run(() => (TException)e);
```
slang25: Use `Task.FromResult` rather than `Task.Run` here

JosephWoodward:
This is great! Thanks so much for fixing this @gmilani ! I'm working towards getting v4 out so this will be in the next beta release in the next few days. Thanks!

Issue #457, opened 2017-08-17
Should.ThrowAsync<T> does seem to work correctly if the underlying Func has a CancellationToken which cancelled


PureKrome:
Hi 👋 

so I'm not sure if this is "by design" and I've not set up my stuff correctly ... or it's a bug.

So I have a simple service I'm testing out. This service accepts a `CancellationToken`. If the service fails to complete in 5 seconds, the CancellationToken cancels, which ends up throwing an exception.

The current code in our unit tests _works_ 👍 

```
var wasCancelled = false;
try
{
    // This hardcoded class is made to throw a `TaskCanceledException` after 5 seconds.
   // The implimentation of this method is a simple `Task.Delay` for 15 seconds.
    await clientService.GetJsonAsync("doesnt-matter");
}
catch (TaskCanceledException)
{
    wasCancelled = true;
}

// Assert.
wasCancelled.ShouldBeTrue();

```

Now, I tried to refactor this into `Shouldly` code like this:

- `var result = await Should.ThrowAsync<TaskCancelledException>(clientService.GetJsonAsync("doesnt-matter"));`
- `var result = await Should.ThrowAsync<TaskCancelledException>(() => clientService.GetJsonAsync("doesnt-matter"));`

and both fail. No exception is thrown by Shouldly.

e.g.

```
Shouldly.ShouldAssertException
Task `clientService.GetJsonAsync("doesnt-matter")`
    should throw
System.Threading.Tasks.TaskCanceledException
    but did not

... <snip >

System.Threading.Tasks.TaskCanceledException
Task is cancelled
  Exception doesn't have a stacktrace
```

---

So I thought I might try and add a test method to the `Shouldly.Tests` projects to see if I can help repo this:

```
[Fact]
public async Task ShouldThrowAWobbly_WhenATaskIsCancelled()
{
    // Arrange.
    // Cancel this calling code after 5 seconds.
    var cancellationTokenSource = new CancellationTokenSource(5);
    var task = Task.Delay(TimeSpan.FromSeconds(10), cancellationTokenSource.Token);

    // Act.
    var result = await Should.ThrowAsync<Exception>(() => task);

    // Assert.
    result.ShouldNotBeNull();
}
```

☝️ So the above code is trying to capture _any_ `Exception` when a Task is cancelled.

Here's a screen shot of the error:

![image](https://user-images.githubusercontent.com/899878/29392978-8c8d176e-8344-11e7-9bdd-98deffea8d37.png)

Interesting:

[...]

Issue #471, opened 2017-10-26
Consider changing Should.ThrowAsync to handle synchronous exceptions and cancellation


paulomorgado:
Something like this:

```C#
try
{
    await actual();
}
catch (TException ex)
{
    return ex;
}
catch (Exception ex)
{
    throw new ShouldAssertException($"Expected exception: {typeof(TException).FullName}{Environment.NewLine}Actual exception: {ex.GetType().FullName}");
}

throw new ShouldAssertException($"Expected exception: {typeof(TException).FullName} but none was thrown.");
```

in the implementation of [Shloud.ThrowAsync](https://github.com/shouldly/shouldly/blob/master/src/Shouldly/ShouldStaticClasses/ShouldThrowTaskAsync.cs#L47-L85), would allow testing for synchronous operations and operation cancelation.

paulomorgado:
knock, knock!

paulomorgado:
knock, knock!

JosephWoodward:
Sounds like a good idea to me. It should at the very least be able to handle cancellation.

paulomorgado:
Do you want me to send a pull request for it? Where would I need to work on?

paulomorgado:
Do you want me to send a pull request for it? Where would I need to work on?

JosephWoodward:
This has now been fixed as part of #560 and will be in the next beta release.

PR #560, opened 2019-03-16
Refactor ShouldThrowAsync for cancellation handing and add test

adamhathcock:
First time to contribute here.

Looking at solving: https://github.com/shouldly/shouldly/issues/457

Might have issues as I couldn't compile all target frameworks on my mac.


adamhathcock:
Also probably solves: https://github.com/shouldly/shouldly/issues/471

JosephWoodward:
Sorry for the late reply, thanks for this!

JosephWoodward:
I'll fix the merge conflict then get it out in the next release.

colinangusmackay:
Is there a time-frame for this to put pushed out to NuGet? The last package that I can see there was November 2018, and this was merged in to Master in June 2019.

JosephWoodward:
Hi @colinangusmackay 

There's currently a [4.0.0 beta](https://www.nuget.org/packages/Shouldly/4.0.0-beta0002) that doesn't introduce any major breaking changes that will include this.

alexnnenn:
I've just tried testing call of function Foo and got different behaviour in test, depending on what variant I use:
```
private static async Task Foo()
{
    await Task.Yield();
    throw new OperationCanceledException();
}

private static async Task ShoudlyVariant2(Type exceptionType)
{
    try
    {
        var exception = await Should.ThrowAsync(Foo, exceptionType);
        Console.WriteLine("ShoudlyV2: " + exception.GetType());
    }
    catch(ShouldAssertException s) when (s.Message.Contains("but did not"))
    {
        Console.WriteLine("Variant 2 did not got " + exceptionType.Name);
    }
}

// manual variant
try
{
    await Foo();
}
catch (Exception x)
{
    Console.WriteLine("Manual:" + x.GetType());
}

// Shoudly variant 1
var exceptionV1 = await Should.ThrowAsync<Exception>(Foo);
Console.WriteLine("ShoudlyV1: " + exceptionV1.GetType());

// Shoudly variant 2
await ShoudlyVariant2(typeof(Exception));
await ShoudlyVariant2(typeof(TaskCanceledException));
await ShoudlyVariant2(typeof(OperationCanceledException));
```

The output is:
```
Manual: System.OperationCanceledException
ShoudlyV1: System.Threading.Tasks.TaskCanceledException
Variant 2 did not got Exception
Variant 2 did not got TaskCanceledException
Variant 2 did not got OperationCanceledException
```

I've expected that It will be OperationCanceledException in all variants. It seems, that ThrowAsync still works wrong in case of canceled task.

I've used latest available release version 4.0.3.

sungam3r:
@alexnnenn Worth to open new issue. This one may be lost.

Issue #559, opened 2019-03-14
ShouldHaveValue() for chainable assertions on nullable values


jnm2:
This a common pattern:

```cs
foo.Bar.ShouldNotBeNull();
foo.Bar.Value.X.ShouldBe(42);
foo.Bar.Value.Name.ShouldBe("bar");
foo.Bar.Value.Date.ShouldBe(Jan3);

baz.Bar.ShouldNotBeNull();
baz.Bar.Value.Date.ShouldBe(Jan4);
```

I would like to be able to remove the `foo.Bar.Value` repetition, which requires a Should method that returns the unpacked value just like `.ShouldHaveSingleItem()` does:

```cs
var bar = foo.Bar.ShouldHaveValue();
bar.X.ShouldBe(42);
bar.Name.ShouldBe("bar");
bar.Date.ShouldBe(Jan3);

baz.Bar.ShouldHaveValue().Date.ShouldBe(Jan4);
```

`.Value` would be fine except that I want a Shouldly error message if it's something that would be expected to break if the test fails.

https://github.com/shouldly/shouldly/issues/271 was closed because the second method would have no advantage over `ShouldBeNull()`.

saiphanish:
I am trying to work on this issue. In which file can I find the above code pattern, in order to fix it?

saiphanish:
@jnm2 ...can you please review my pull request

jnm2:
@saiphanish I'm sorry, I missed this. I'm happy to review but I'm just a contributor like you, so we'll need @JosephWoodward or @slang25's review as well.

SimonCropp:
@jnm2 now u have commit access. do u want to revisit this?

jnm2:
Yes, I would like to. I'll get a PR ready and then close #559 since it's based on a very old version and works with `List<T>` instead of `Nullable<T>`.

jnm2:
Actually, now that we already have separate overloads of ShouldNotBeNull for reference types and value types, I'd like to propose that both of those return non-nullable T (unwrapping System.Nullable in the case of value types). This will kill two birds with one stone because I've also wanted ShouldNotBeNull to be chainable for reference types. It's also a super simple change. PR incoming.

PR #563, opened 2019-04-07
#559 Added ShouldHaveValue()

saiphanish:
ShouldHaveValue() for chainable assertions on nullable values #559


jnm2:
I'd like to take the design in a different direction (#684) than I originally proposed. Your willingness to help means a lot even though it didn't work out this time; thank you!

PR #567, opened 2019-05-04
Reference new full framework reference assemblies on NuGet

JosephWoodward:


src/Shouldly/Shouldly.csproj
```diff
@@ -21,7 +20,7 @@
     <Optimize>false</Optimize>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFramework)'== 'net451'">
-      <FrameworkPathOverride>$(NuGetPackageRoot)microsoft.targetingpack.netframework.v4.5.1/1.0.1/lib/net451/</FrameworkPathOverride>
+      <FrameworkPathOverride>$(NuGetPackageRoot)microsoft.netframework.referenceassemblies.net451/1.0.0-preview.1/build/.NETFramework/v4.5.1</FrameworkPathOverride>
```
slang25: This property can go now, as it is disabled explicitly by the ReferenceAssemblies package

src/Shouldly/Shouldly.csproj
```diff
@@ -34,7 +33,7 @@
   <ItemGroup Condition=" '$(TargetFramework)' == 'net451' ">
     <Reference Include="System" />
     <Reference Include="Microsoft.CSharp" />
-    <PackageReference Include="Microsoft.TargetingPack.NETFramework.v4.5.1" Version="1.0.1" ExcludeAssets="All" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies.net451" Version="1.0.0-preview.1" PrivateAssets="All" />
```
slang25: You can and probably should omit the `.net451` as it's intended to be referenced by the meta-package, the way NuGet works with meta-packages means that only `net451` will be resolved and downloaded.

src/Shouldly/Shouldly.csproj
JosephWoodward: I did wonder this, but it blows up if I remove it.

slang25: Interesting, was that locally or on CI?

slang25: Just tried it and it works here. I did a write-up on how it works:
https://stu.dev/dotnet-framework-reference-assemblies-via-nuget/
`FrameworkPathOverride` should do nothing when this package is referenced.

JosephWoodward: > Interesting, was that locally or on CI?

Weird, tried it and it worked for some reason.

Issue #575, opened 2019-08-22
Return the input value from ShouldNotBeNull


stop-cran:
For me it's a common case when I check if a value is null only to correctly perform checks on its property:

```
var value = GetValue();

value.ShouldNotBeNull();

var prop = value.SomeProperty;

// further tests for prop
```

Up to me it's convenient to chain these checks, like it's done for `ShouldBeOfType` - simply returning the input value as a result of `ShouldNotBeNull`:

```
GetValue()
    .ShouldNotBeNull()
    .SomeProperty
    // further shoulds
```

So I'd propose following signatures:

```
[Shouldly.ShouldlyMethodsAttribute()]
public class static ShouldBeNullExtensions
{
    public static T ShouldNotBeNull<T>(this T actual) { }
    public static T ShouldNotBeNull<T>(this T actual, string customMessage) { }
    public static T ShouldNotBeNull<T>(this T actual, System.Func<string> customMessage) { }

    // ... other methods
}
```

Jehoel:
Duplicate of https://github.com/shouldly/shouldly/issues/61

gambrose:
I do not think this is a duplicate of #61 as there are other use cases than fluid chaining.

In the example given:
```
var value = GetValue();

value.ShouldNotBeNull();

var prop = value.SomeProperty;
```
`var` would be nullable ref type which would generate compiler warning for `value.SomeProperty` if nullable ref checks enabled.

Where as here var would be inferred as a non null ref type.
```
var value = GetValue().ShouldNotBeNull();

var prop = value.SomeProperty;
```

JosephWoodward:
I'm going to close this off as one of Shouldy's aims is to be simple and we've actively tried to keep a fluent API out of it.

PR #576, opened 2019-08-26
#575 Return the input value from ShouldNotBeNull.

stop-cran:
Amended existing methods and tests.


sungam3r:
Fixes #575

JosephWoodward:
Thank you so much for your pull request but one of Shouldly's design goals is for a simple, consistent API. We've always been keen to keep fluent APIs out of it as there are other libraries that support such a functionality.

PR #581, opened 2019-10-14
Nullable reference types

### LINKED ISSUE #579: Provide annotations for C# 8 Nullable Reference Types
I'm starting to run into scenarios where I'd appreciate annotations on Shouldly. Would you accept a PR to move the LangVersion to 8.0 (necessary to annotate reference types with `?` in public APIs) and enable https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references? This does not require targeting .NET Core 3.0.

```cs
var foo = ...;
foo.ShouldNotBeNull();

// I'd like to be able to remove the `!`
foo!.Bar.ShouldBe(42);
```

In order to let the compiler know that `ShouldNotBeNull` means that `foo` can't be null lower down, you'd add `[NotNull]` to this line:

https://github.com/shouldly/shouldly/blob/4f990555884d60e07bc1278040a131a80e9f5031/src/Shouldly/ShouldlyExtensionMethods/ShouldBeNullExtensions.cs#L31

```diff
-   public static void ShouldNotBeNull<T>(this T actual) 
+   public static void ShouldNotBeNull<T>([NotNull] this T actual) 
```

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullattribute:

> Specifies that an output is not `null` even if the corresponding type allows it.

Even though the the documentation refers to it as an output, and `actual` above is a by-value parameter, I have verified that `[NotNull]` has the intended effect.

It's also been really fun using https://github.com/tunnelvisionlabs/ReferenceAssemblyAnnotator for two benefits, so I'd recommend it:

- It defines the System.Diagnostics.CodeAnalysis nullability attributes as internal attributes, which you would otherwise have to do manually.
- It annotates the .NET Framework and Standard reference assemblies by copying annotations from matching APIs in .NET Core 3.0 so that your codebase can benefit internally.

jnm2:
Closes #579.

I tried to keep the big initial annotation commit as straightforward as possible, saving more complex fixes for later commits. For files where the changes aren't absolutely obvious, please consider reviewing single commits at a time.

➡ Please comment if you'd like any of the types in these APIs to be annotated differently. I simply annotated them as nullable if `null` was tolerated in previous releases and left them non-nullable if `null` would have caused an exception. All nullable annotations (`?` and attributes) are a description of my understanding of the past rather than a statement from me about the future.

### DIFF
--- a/src/DocumentationExamples/DocumentationExamples.csproj
+++ b/src/DocumentationExamples/DocumentationExamples.csproj
@@ -18,7 +18,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1" />
     <PackageReference Include="xunit" Version="2.4.1" />
-    <PackageReference Include="PublicApiGenerator" Version="10.0.0-beta3" />
+    <PackageReference Include="PublicApiGenerator" Version="10.0.0" />
     <PackageReference Include="TestStack.ConventionTests" Version="3.0.1" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.9.0" />
   </ItemGroup>
--- a/src/DocumentationExamples/DocumentationExamples.csproj
+++ b/src/DocumentationExamples/DocumentationExamples.csproj
@@ -18,7 +18,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1" />
     <PackageReference Include="xunit" Version="2.4.1" />
-    <PackageReference Include="PublicApiGenerator" Version="8.1.0" />
+    <PackageReference Include="PublicApiGenerator" Version="10.0.0-beta3" />
     <PackageReference Include="TestStack.ConventionTests" Version="3.0.1" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.9.0" />
   </ItemGroup>
--- a/src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
+++ b/src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
@@ -9,139 +9,139 @@ namespace Shouldly
     public static class DynamicShould
     {
         public static void HaveProperty([System.Runtime.CompilerServices.Dynamic] object dynamicTestObject, string propertyName) { }
-        public static void HaveProperty([System.Runtime.CompilerServices.Dynamic] object dynamicTestObject, string propertyName, string? customMessage) { }
         public static void HaveProperty([System.Runtime.CompilerServices.Dynamic] object dynamicTestObject, string propertyName, System.Func<string?>? customMessage) { }
+        public static void HaveProperty([System.Runtime.CompilerServices.Dynamic] object dynamicTestObject, string propertyName, string? customMessage) { }
     }
     [Shouldly.ShouldlyMethods]
     public static class ObjectGraphTestExtensions
     {
         public static void ShouldBeEquivalentTo([System.Diagnostics.CodeAnalysis.NotNullIfNotNull("expected")] this object? actual, [System.Diagnostics.CodeAnalysis.NotNullIfNotNull("actual")] object? expected) { }
-        public static void ShouldBeEquivalentTo([System.Diagnostics.CodeAnalysis.NotNullIfNotNull("expected")] this object? actual, [System.Diagnostics.CodeAnalysis.NotNullIfNotNull("actual")] object? expected, string? customMessage) { }
         public static void ShouldBeEquivalentTo([System.Diagnostics.CodeAnalysis.NotNullIfNotNull("expected")] this object? actual, [System.Diagnostics.CodeAnalysis.NotNullIfNotNull("actual")] object? expected, System.Func<string?>? customMessage) { }
+        public static void ShouldBeEquivalentTo([System.Diagnostics.CodeAnalysis.NotNullIfNotNull("expected")] this object? actual, [System.Diagnostics.CodeAnalysis.NotNullIfNotNull("actual")] object? expected, string? customMessage) { }
     }
     [Shouldly.ShouldlyMethods]
     public static class Should
     {
         public static void CompleteIn(System.Action action, System.TimeSpan timeout) { }
-        public static void CompleteIn(System.Action action, System.TimeSpan timeout, string? customMessage) { }
-        public static void CompleteIn(System.Action action, System.TimeSpan timeout, System.Func<string?>? customMessage) { }
-        public static T CompleteIn<T>(System.Func<T> function, System.TimeSpan timeout) { }
-        public static T CompleteIn<T>(System.Func<T> function, System.TimeSpan timeout, string? customMessage) { }
-        public static T CompleteIn<T>(System.Func<T> function, System.TimeSpan timeout, System.Func<string?>? customMessage) { }
         public static void CompleteIn(System.Func<System.Threading.Tasks.Task> actual, System.TimeSpan timeout) { }
-        public static void CompleteIn(System.Func<System.Threading.Tasks.Task> actual, System.TimeSpan timeout, string? customMessage) { }
-        public static void CompleteIn(System.Func<System.Threading.Tasks.Task> actual, System.TimeSpan timeout, System.Func<string?>? customMessage) { }
-        public static T CompleteIn<T>(System.Func<System.Threading.Tasks.Task<T>> actual, System.TimeSpan timeout) { }
-        public static T CompleteIn<T>(System.Func<System.Threading.Tasks.Task<T>> actual, System.TimeSpan timeout, string? customMessage) { }
-        public static T CompleteIn<T>(System.Func<System.Threading.Tasks.Task<T>> actual, System.TimeSpan timeout, System.Func<string?>? customMessage) { }
         public static void CompleteIn(System.Threading.Tasks.Task actual, System.TimeSpan timeout) { }
-        public static void CompleteIn(System.Threading.Tasks.Task actual, System.TimeSpan timeout, string? customMessage) { }
+        public static void CompleteIn(System.Action action, System.TimeSpan timeout, System.Func<string?>? customMessage) { }
+        public static void CompleteIn(System.Action action, System.TimeSpan timeout, string? customMessage) { }
+        public static void CompleteIn(System.Func<System.Threading.Tasks.Task> actual, System.TimeSpan timeout, System.Func<string?>? customMessage) { }
+        public static void CompleteIn(System.Func<System.Threading.Tasks.Task> actual, System.TimeSpan timeout, string? customMessage) { }
         public static void CompleteIn(System.Threading.Tasks.Task actual, System.TimeSpan timeout, System.Func<string?>? customMessage) { }
+        public static void CompleteIn(System.Threading.Tasks.Task actual, System.TimeSpan timeout, string? customMessage) { }
+        public static T CompleteIn<T>(System.Func<System.Threading.Tasks.Task<T>> actual, System.TimeSpan timeout) { }
+        public static T CompleteIn<T>(System.Func<T> function, System.TimeSpan timeout) { }
         public static T CompleteIn<T>(System.Threading.Tasks.Task<T> actual, System.TimeSpan timeout) { }
-        public static T CompleteIn<T>(System.Threading.Tasks.Task<T> actual, System.TimeSpan timeout, string? customMessage) { }
+        public static T CompleteIn<T>(System.Func<System.Threading.Tasks.Task<T>> actual, System.TimeSpan timeout, System.Func<string?>? customMessage) { }
+        public static T CompleteIn<T>(System.Func<System.Threading.Tasks.Task<T>> actual, System.TimeSpan timeout, string? customMessage) { }
+        public static T CompleteIn<T>(System.Func<T> function, System.TimeSpan timeout, System.Func<string?>? customMessage) { }
+        public static T CompleteIn<T>(System.Func<T> function, System.TimeSpan timeout, string? customMessage) { }
         public static T CompleteIn<T>(System.Threading.Tasks.Task<T> actual, System.TimeSpan timeout, System.Func<string?>? customMessage) { }
+        public static T CompleteIn<T>(System.Threading.Tasks.Task<T> actual, System.TimeSpan timeout, string? customMessage) { }
         public static void NotThrow(System.Action action) { }
-        public static void NotThrow(System.Action action, string? customMessage) { }
-        public static void NotThrow(System.Action action, System.Func<string?>? customMessage) { }
-        public static T NotThrow<T>(System.Func<T> action) { }
-        public static T NotThrow<T>(System.Func<T> action, string? customMessage) { }
-        public static T NotThrow<T>(System.Func<T> action, System.Func<string?>? customMessage) { }
-        public static void NotThrow(System.Threading.Tasks.Task action) { }
-        public static void NotThrow(System.Threading.Tasks.Task action, string? customMessage) { }
-        public static void NotThrow(System.Threading.Tasks.Task action, System.Func<string?>? customMessage) { }
-        public static T NotThrow<T>(System.Threading.Tasks.Task<T> action) { }
-        public static T NotThrow<T>(System.Threading.Tasks.Task<T> action, string? customMessage) { }
-        public static T NotThrow<T>(System.Threading.Tasks.Task<T> action, System.Func<string?>? customMessage) { }
         public static void NotThrow(System.Func<System.Threading.Tasks.Task> action) { }
-        public static void NotThrow(System.Func<System.Threading.Tasks.Task> action, string? customMessage) { }
+        public static void NotThrow(System.Threading.Tasks.Task action) { }
+        public static void NotThrow(System.Action action, System.Func<string?>? customMessage) { }
+        public static void NotThrow(System.Action action, string? customMessage) { }
         public static void NotThrow(System.Func<System.Threading.Tasks.Task> action, System.Func<string?>? customMessage) { }
-        public static void NotThrow(System.Threading.Tasks.Task action, System.TimeSpan timeoutAfter) { }
-        public static void NotThrow(System.Threading.Tasks.Task action, System.TimeSpan timeoutAfter, string? customMessage) { }
-        public static void NotThrow(System.Threading.Tasks.Task action, System.TimeSpan timeoutAfter, System.Func<string?>? customMessage) { }
         public static void NotThrow(System.Func<System.Threading.Tasks.Task> action, System.TimeSpan timeoutAfter) { }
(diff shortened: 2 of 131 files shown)


src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
```diff
@@ -7,156 +7,175 @@ public enum Case
     }
     [System.ObsoleteAttribute("This class is only kept here for backwards compatibility. Please use ShouldAssert" +
         "Exception instead.")]
+    [System.Runtime.CompilerServices.NullableAttribute(0)]
     public class ChuckedAWobbly : System.Exception
```
jnm2: @JosephWoodward This diff is going to look miserable until https://github.com/ApiApprover/ApiApprover/issues/54 is fixed. I don't know if that means this PR should wait?

JosephWoodward: @jnm2 Shouldly is currently in a beta release of a new major version change so if we were going to introduce any breaking changes to the public API then now is a suitable time to do it (providing it didn't break backwards compatibility of course).

jnm2: Cool. I don't think any of these API changes are breaking changes. (Unless I unintentionally changed something besides NRT annotations.) If a project already has `<Nullable>enable</Nullable>` and upgrades to the annotated Shouldly, new warnings might show up now that the compiler knows for sure which inputs and outputs from Shouldly can and can't be null. I believe more of .NET Core is receiving annotations in 3.1, not a major version change.

src/Shouldly.Tests/Shouldly.Tests.csproj
```diff
@@ -17,9 +15,8 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1" />
-    <PackageReference Include="PublicApiGenerator" Version="8.0.1" />
+    <PackageReference Include="PublicApiGenerator" Version="10.0.0-beta1" />
     <PackageReference Include="TestStack.ConventionTests" Version="3.0.1" />
-    <DotNetCliToolReference Include="dotnet-xunit" Version="2.3.1" />
```
jnm2: `dotnet-test` was discontinued: `https://github.com/xunit/xunit/issues/1835#issuecomment-431199286`

Version 2.3.1 has no rollforward policy and can only run on .NET Core 2.0 which is EOL.

src/Shouldly.Tests/Shouldly.Tests.csproj
```diff
@@ -2,7 +2,6 @@
   <PropertyGroup>
     <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);net472;</TargetFrameworks>
-    <LangVersion>latest</LangVersion>
```
jnm2: Moved to Directory.Build.props so that it could be next to `<Nullable>enable</Nullable>`

src/Shouldly/Internals/EqualityComparer.cs
```diff
@@ -33,10 +33,10 @@ public bool Equals(T x, T y)
             // Null?
             if (!type.IsValueType() || (type.IsGenericType() && type.GetGenericTypeDefinition().IsAssignableFrom(NullableType)))
             {
-                if (object.Equals(x, default(T)))
-                    return object.Equals(y, default(T));
+                if (object.Equals(x, null))
+                    return object.Equals(y, null);
 
-                if (object.Equals(y, default(T)))
+                if (object.Equals(y, null))
```
jnm2: `default(T)` is always `null` given the `if` condition. This is nicer than doing `default(T)!`.

src/Shouldly/Internals/EqualityComparer.cs
```diff
@@ -33,10 +33,10 @@ public bool Equals(T x, T y)
             // Null?
             if (!type.IsValueType() || (type.IsGenericType() && type.GetGenericTypeDefinition().IsAssignableFrom(NullableType)))
```
jnm2: This check would be simpler as `default(T)! is null`.

src/Shouldly/Internals/EqualityComparerAdapter.cs
```diff
@@ -18,9 +18,9 @@ public EqualityComparerAdapter(IEqualityComparer<T> innerComparer)
             _innerComparer = innerComparer;
         }
 
-        public new bool Equals(object x, object y)
+        public new bool Equals(object? x, object? y)
         {
-            return _innerComparer.Equals((T) x, (T) y);
+            return _innerComparer.Equals((T)x!, (T)y!);
```
jnm2: The cast will always succeed when `T` is a nullable type at runtime (either a reference type or `Nullable<>`).

src/Shouldly/Internals/Is.cs
```diff
@@ -3,18 +3,18 @@
...
 using System.Text.RegularExpressions;
-using System.Reflection;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Shouldly
 {
     internal static class Is
     {
-        public static bool InRange<T>(T comparable, T @from, T to) where T : IComparable<T>
+        public static bool InRange<T>([DisallowNull] T comparable, [AllowNull] T @from, [AllowNull] T to) where T : IComparable<T>
```
jnm2: `[AllowNull]` and `[DisallowNull]` are the only way to annotate these parameters since C# still does not allow `T?` here. (Allowing `T?` without a class or struct constraint would be a CLR change.)

src/Shouldly/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
```diff
@@ -102,62 +103,62 @@ public static void ShouldAllBe<T>(this IEnumerable<T> actual, [InstantHandle] Ex
...
+        public static void ShouldAllBe<T>(this IEnumerable<T> actual, [InstantHandle] Expression<Func<T, bool>> elementPredicate, [InstantHandle] Func<string?>? customMessage)
         {
             var condition = elementPredicate.Compile();
             var actualResults = actual.Where(part => !condition(part));
             if (actualResults.Any())
                 throw new ShouldAssertException(new ActualFilteredWithPredicateShouldlyMessage(elementPredicate.Body, actualResults, actual, customMessage).ToString());
         }
 
-        public static void ShouldBeEmpty<T>(this IEnumerable<T> actual)
+        public static void ShouldBeEmpty<T>([NotNull] this IEnumerable<T>? actual)
```
jnm2: `[NotNull]` initially seems to conflict with the `?`, but there is no conflict. `[NotNull]` is indicating that if this call returns, the expression that had been passed to `actual` cannot have been null. At the same time, the method accepts null inputs without a compile-time warning.

You could also choose to go with `ShouldBeEmpty<T>(this IEnumerable<T> actual)` in which case the compiler would warn if the user does `foo.ShouldBeEmpty();` without first proving to the compiler that `foo` can't be null (or using `!`).

src/Shouldly/Internals/Numerics.cs
```diff
@@ -14,7 +15,7 @@ internal static class Numerics
         /// </summary>
         /// <param name="obj">The object to check</param>
         /// <returns>true if the object is a numeric type</returns>
-        public static bool IsNumericType(Object obj)
+        public static bool IsNumericType([NotNullWhen(true)] object? obj)
```
jnm2: `NotNullWhen` is used to alert the compiler's flow state analysis to this situation:

```cs
nullableThing = ...
if (IsNumericType(nullableThing))
{
    nullableThing.GetHashCode(); // No warning because IsNumericType returned true
}

nullableThing.GetHashCode(); // Compiler warning because `nullableThing` could be null at this point
```

src/Shouldly/Internals/ShouldThrowAssertionContext.cs
```diff
@@ -4,18 +4,18 @@ namespace Shouldly
...
         /// <summary>
         /// Only pass stacktrace if asynchronous
         /// </summary>
-        internal ShouldThrowAssertionContext(object expected, object actual = null, string exceptionMessage = null,
+        internal ShouldThrowAssertionContext(object? expected, object? actual = null, string? exceptionMessage = null,
             bool isAsync = false,
-            System.Diagnostics.StackTrace stackTrace = null,
-            [CallerMemberName] string shouldlyMethod = null) : base(shouldlyMethod, expected, actual, stackTrace)
+            System.Diagnostics.StackTrace? stackTrace = null,
+            [CallerMemberName] string shouldlyMethod = null!) : base(shouldlyMethod, expected, actual, stackTrace)
```
jnm2: I don't think there's a situation where the default value `null` could be used.
https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute#remarks

src/Shouldly/Internals/SourceCodeTextGetter.cs
```diff
@@ -84,7 +84,7 @@ string GetCodePart()
             {
                 var codeLines = string.Join("\n", File.ReadAllLines(FileName).Skip(LineNumber).ToArray());
 
-                var indexOf = codeLines.IndexOf(_shouldMethod);
+                var indexOf = codeLines.IndexOf(_shouldMethod!);
```
jnm2: The way to avoid all these dammit operators (`!`) would be to pass the value between methods rather than making it a class field. The state of a mutable class field is also harder to reason about between methods for people as well as compilers, so I would recommend this.

src/Shouldly/Internals/StringHelpers.cs
```diff
@@ -170,24 +170,24 @@ internal static string ToSafeString(this char c)
...
+        static string CommaDelimited<T>(this IEnumerable<T> enumerable) where T : class?
         {
             return enumerable.DelimitWith(", ");
         }
 
-        static string DelimitWith<T>(this IEnumerable<T> enumerable, string separator) where T : class
+        static string DelimitWith<T>(this IEnumerable<T> enumerable, string? separator) where T : class?
         {
-            return string.Join(separator, enumerable.Select(i => Equals(i, default(T)) ? null : i.ToString()).ToArray());
+            return string.Join(separator, enumerable.Select(i => Equals(i, null) ? null : i.ToString()).ToArray());
```
jnm2: `default(T)` was always `null` already. Using `null` is even nicer now that the other option is `default(T)!`.

src/Shouldly/MessageGenerators/DictionaryShouldContainKeyAndValueMessageGenerator.cs
```diff
@@ -14,6 +15,9 @@ public override bool CanProcess(IShouldlyAssertionContext context)
 
         public override string GenerateErrorMessage(IShouldlyAssertionContext context)
         {
+            Debug.Assert(context.Actual is IDictionary);
+            Debug.Assert(context.Key is object);
```
jnm2: ⚠ Not sure how you'd like to handle these. This method is relying on the fragile assumption that remote code will have set these fields, and these asserts silence the warnings this causes now that NRTs are enabled. The way I would handle it would be to require these values to be passed to the DictionaryShouldContainKeyAndValueMessageGenerator constructor.

src/Shouldly/ShouldlyExtensionMethods/ShouldBe/GenericShouldBeTestExtensions.cs
```diff
@@ -10,53 +11,72 @@ namespace Shouldly
...
-            if (ShouldlyConfiguration.CompareAsObjectTypes.Contains(typeof(T).FullName) || typeof(T) == typeof(string))
+            if (ShouldlyConfiguration.CompareAsObjectTypes.Contains(typeof(T).FullName!) || typeof(T) == typeof(string))
                 actual.AssertAwesomely(v => Is.Equal(v, expected, new ObjectEqualityComparer<T>()), actual, expected, customMessage);
             else
                 actual.AssertAwesomely(v => Is.Equal(v, expected), actual, expected, customMessage);
         }
 
         [ContractAnnotation("actual:null,expected:null => halt")]
-        public static void ShouldNotBe<T>(this T actual, T expected)
+        public static void ShouldNotBe<T>([AllowNull] this T actual, [AllowNull] T expected)
```
jnm2: `[AllowNull]` is needed on each so that generic inference doesn't get upset when one expression is nullable and the other isn't.

src/Shouldly/ShouldlyExtensionMethods/ShouldBe/StringNullOrEmptyTestExtensions.cs
```diff
@@ -1,43 +1,44 @@
 ﻿using System;
 using JetBrains.Annotations;
+using NotNullAttribute = System.Diagnostics.CodeAnalysis.NotNullAttribute;
 
 namespace Shouldly
 {
     public static partial class ShouldBeStringTestExtensions
     {
         [ContractAnnotation("actual:notnull => halt")]
```
jnm2: Side note: I think these JetBrains annotations might actually be incorrect. If actual is `""`, a non-null value, the method does not halt.

src/Shouldly/ShouldlyExtensionMethods/ShouldBeBooleanExtensions.cs
```diff
@@ -8,33 +9,33 @@ namespace Shouldly
     [ShouldlyMethods]
     public static class ShouldBeBooleanExtensions
     {
-        public static void ShouldBeTrue(this bool actual)
+        public static void ShouldBeTrue([DoesNotReturnIf(false)] this bool actual)
```
jnm2: `DoesNotReturnIf` is the annotation that powers scenario like this:

```cs
nullableThing.GetHashCode(); // Compiler warning, possible NRE
(nullableThing != null).ShouldBeTrue();
nullableThing.GetHashCode(); // No warning because ShouldBeTrue did not throw an exception
```

src/Shouldly/ShouldlyExtensionMethods/ShouldBeDictionaryTestExtensions.cs
```diff
@@ -3,71 +3,83 @@
...
-namespace Shouldly 
+namespace Shouldly
 {
     [DebuggerStepThrough]
     [ShouldlyMethods]
-    public static class ShouldBeDictionaryTestExtensions 
+    public static class ShouldBeDictionaryTestExtensions
     {
         public static void ShouldContainKey<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key)
+            where TKey : notnull
```
jnm2: This is necessary in order to use `TKey` for `IDictionary<TKey, TValue>` due to the generic constraint added to `IDictionary<,>` in the BCL.

src/Shouldly/ShouldlyExtensionMethods/ShouldBeNullExtensions.cs
```diff
@@ -9,38 +11,38 @@ namespace Shouldly
     public static class ShouldBeNullExtensions
     {
         [ContractAnnotation("actual:notnull => halt")]
-        public static void ShouldBeNull<T>(this T actual)
+        public static void ShouldBeNull<T>([MaybeNull] this T actual)
```
jnm2: If this method returns, the `actual` expression is definitely null. `[MaybeNull]` is the strongest the .NET annotations get. Threw it in for completeness.

```cs
var nonnullableThing = new object();
nonnullableThing.GetHashCode(); // No warning
nonnullableThing.ShouldBeNull();
nonnullableThing.GetHashCode(); // Warning that nonnullableThing may be null
```

src/Shouldly/ShouldlyMessage.cs
```diff
@@ -134,7 +134,7 @@ internal class ExpectedOrderShouldlyMessage : ShouldlyMessage
 
     internal class ExpectedEquvalenceShouldlyMessage : ShouldlyMessage
```
jnm2: This is a typo.

danielmarbach:
Beta 3 of PublicApiGenerator is on nuget

jnm2:
@JosephWoodward Everything is a go as soon as you get a chance to review, and there's still the GitVersion failure on the Ubuntu in the master branch which is what the CI failure is about.

Is there anyone else I should ping for review? The more eyes the better for something like this, probably. :D

src/DocumentationExamples/DocumentationExamples.csproj
```diff
@@ -17,7 +17,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1" />
     <PackageReference Include="xunit" Version="2.4.1" />
-    <PackageReference Include="PublicApiGenerator" Version="8.1.0" />
+    <PackageReference Include="PublicApiGenerator" Version="10.0.0-beta3" />
```
sungam3r: 10.0.0-beta3 has a bug with `readonly` structs, it was fixed in beta4

jnm2: This doesn't affect Shouldly, but I added a commit updating to 10.0.0.

jnm2:
@JosephWoodward Ready for review.

JosephWoodward:
@jnm2 Amazing work :D I'll try and go through this over the next few days. Great PR though!

jnm2:
@JosephWoodward New force push. Two of the commits have already been merged in https://github.com/shouldly/shouldly/pull/596 and I fixed merge conflicts caused by https://github.com/shouldly/shouldly/pull/595.

JosephWoodward:
@jnm2 Nice! So is this good to go? (pending review of course)

jnm2:
@JosephWoodward Yes!

jnm2:
@JosephWoodward I merged with master to fix merge commits, then I recognized the build failure as being related to certain versions of Mono and Cake pre-0.37 and fixed that, then I fixed some new warnings that will crop up in VS 16.5.

I'm not sure reviewing and merging will get more appealing as time goes on, and I don't know what to do to make that easier.

jnm2:
Pushing a merge commit to resolve conflicts soon.

jnm2:
@SimonCropp I was going to push a merge commit to resolve the conflicts. Would it make more sense to wait till the dust settles from the PRs you're merging, or would there be another approach altogether that would work better for you?

SimonCropp:
i can hold off for a few hours. i i can handle the conflicts and merge for u

jnm2:
@SimonCropp You might have better context than me for fixing the conflicts since you are familiar with what's happened on master in the meantime. That sounds best, but I'm also happy to do it. Once this PR merges, there is a good chance that most outstanding PRs will have merge conflicts too. The tradeoff is fixing things individually versus adding the equivalent work in bulk to this PR. Is one better than the other?

SimonCropp:
@jnm2 i manually merged here https://github.com/shouldly/shouldly/pull/634. there still are a few nullability warnings. i was wondering if you could have a look at the warnings on master, given you clearly know more about nullability  than me

jnm2:
@SimonCropp For sure!

Issue #509, opened 2018-07-06
ShouldAllBeEqual for enumerables


jnm2:
If you want to assert that no two values differ in an enumerable, it looks a little round-about:

```cs
enumerable.Distinct().ShouldHaveSingleItem();
```

And the message isn't as direct:

```
    should have single item but had
2
    items and was
[Color [A=255, R=32, G=31, B=53], Color [Empty]]
```

What would you think of `enumerable.ShouldAllBeEqual()` which would return the value just like `ShouldHaveSingleItem`? The message could then be something like this:

```
    should have all items equal but had:
1 occurrence of [Color [A=255, R=32, G=31, B=53]
2 occurrences of [Color [Empty]]
```

Haneen96:
Hey All,
How can i test my added methods to Shouldly?

jnm2:
I would start by copying the existing test pattern, e.g. https://github.com/shouldly/shouldly/blob/master/src/Shouldly.Tests/ShouldBeUnique/IntegerArrayScenario.cs.

Haneen96:
Hello,,
When I test the my added method ( ShouldAllBeEqual() ), it goes to ( ShouldAllBeMessageGenerator.cs "previous class were in Shouldly" ). Do you have any idea how can I make it go to ( ShouldAllBeEqualMessageGenerator.cs "I made to output the needed message") ??

jnm2:
@Haneen96 Oh dear. This is because the regexes in all the message generators are not specific enough. Look at this:

https://github.com/shouldly/shouldly/blob/3574211043d8fae153d11ee8ad0d9dfd92017d6f/src/Shouldly/MessageGenerators/ShouldAllBeMessageGenerator.cs#L7-L12

What's happening is that the method name `ShouldAllBeEqual` is matching the regex `ShouldAllBe` and `ShouldAllBeMessageGenerator` is higher in the list than `ShouldAllBeEqualMessageGenerator`, so it gets priority. One fix would be to put it lower in the list, but an even better fix would be to fix the larger matching problem in all the generators.

The regexes should probably all start and end with `\b` (word boundary). @JosephWoodward Would you agree?

PR #588, opened 2019-10-31
ShouldAllBeEqual for enumerables #509

Majdi-web:
**### We use the ShouldBe statement to catch errors easier and to show you the differences between the two collections.**

### DIFF
--- a/src/Shouldly.Tests/ShouldBe/WithTolerance/DateTimeScenario.cs
+++ b/src/Shouldly.Tests/ShouldBe/WithTolerance/DateTimeScenario.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using Shouldly.Tests.Strings;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


src/Shouldly.Tests/ShouldBe/WithTolerance/DateTimeScenario.cs
```diff
@@ -13,72 +13,91 @@ public void DateTimeScenarioShouldFail()
             var dateString = date.ToString("o");
             var expected = new DateTime(2000, 6, 1, 1, 0, 1);
             var expectedString = expected.ToString("o");
-            Verify.ShouldFail(() =>
```
SimonCropp, line 16: this is this deleted?

SimonCropp:
the diff on this looks really strange. also there is a lot of whitespace. does it look correct to you?

PR #595, opened 2019-12-01
Remove obsoleted types and members

### LINKED ISSUE #590: Remove the obsoleted ChuckedAWobbly
It's been obsolete for a while. What would be a good date by which it could be removed?

jnm2:
Closes #590


JosephWoodward:
Good idea, might as well do this now we're introducing a major version.

PR #603, opened 2020-03-04
Optimized `.ShouldBeUnique()` for large datasets

### LINKED ISSUE #602: `.ShouldBeUnique()` runs in O(n!) time
Had a test the other day upon ~500,000 record dataset (I've since refactored this), but attempting to use `.ShouldBeUnique()` upon this dataset resulted in a test that never completed (not in the amount of time I was willing to wait, anyway.

The offending code can be seen [here](https://github.com/shouldly/shouldly/blob/d0f7ffbfb34a4202d2d2f6a3cadc4f519480a402/src/Shouldly/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs#L310).

```cs
private static List<object> GetDuplicates<T>(IEnumerable<T> items)
{
    var list = new List<object>();
    var duplicates = new List<object>();

    foreach (object o1 in items)
    {
        duplicates.AddRange(list.Where(o2 => o1 != null && o1.Equals(o2)));
        list.Add(o1);
    }

    return duplicates;
}
```

JakenVeina:
With corresponding test.

Resolves #602.

### DIFF
--- a/src/Shouldly.Tests/ShouldBeUnique/LargeDatasetScenario.cs
+++ b/src/Shouldly.Tests/ShouldBeUnique/LargeDatasetScenario.cs
@@ -1,0 +1,18 @@
+﻿using System.Linq;
+
+using Xunit;
+
+namespace Shouldly.Tests.ShouldBeUnique
+{
+    public class LargeDatasetScenario
+    {
+        [Fact]
+        public void ShouldPass()
+        {
+            Enumerable.Range(1, 500000)
+                .ToArray()
+                .ShouldBeUnique();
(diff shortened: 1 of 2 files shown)


JosephWoodward:
Oh nice, thanks for this!

JakenVeina:
Not a problem. Not sure why the build is failing, but I'm fairly certain it's not related to the changes. It definitely builds and tests fine locally.

mrCraick:
This will not work for classes that do not override the Equals() and GetHashCode() functions.

PR #605, opened 2020-03-10
Fix CI so that pull requests can be merged

jnm2:
@JosephWoodward I cherry-picked this from https://github.com/shouldly/shouldly/pull/581 because @hrai was asking for help with a PR build failure.


JosephWoodward:
Oh great, thank for that.

PR #610, opened 2020-03-27
ShouldBe does a sequence comparison for ReadOnlyMemory

### LINKED ISSUE #562: ShouldBe does not do a sequence comparison for ReadOnlyMemory
Fails (compares reference equality):

```cs
ReadOnlyMemory<byte> foo = new byte[] { 1, 2, 3 }.AsMemory();

foo.ShouldBe(new byte[] { 1, 2, 3 });
```

Workaround:

```cs
ReadOnlyMemory<byte> foo = new byte[] { 1, 2, 3 }.AsMemory();

MemoryMarshal.ToEnumerable(foo).ShouldBe(new byte[] { 1, 2, 3 });
```

It would be more intuitive if Shouldly did the `MemoryMarshal.ToEnumerable` call (using reflection in .NET Standard builds).

hrai:
Closes #562 

@jnm2 I've created this as there was no update on the linked issue so I assumed no one was working on the fix.


src/Shouldly/Internals/TypeExtensions.cs
```diff
@@ -22,5 +22,29 @@ internal static class TypeExtensions
         public static bool IsDefined(this Type type, Type attributeType, bool inherit) =>
             type.GetTypeInfo().IsDefined(attributeType, inherit);
 #endif
+
+        public static bool IsMemory(this Type type, out Type elementType)
+        {
+            if (type.IsGenericType() && type.GetGenericTypeDefinition()?.FullName == "System.Memory`1")
```
sungam3r:
```suggestion
            if (type.IsGenericType() && type.GetGenericTypeDefinition().FullName == "System.Memory`1")
```

sungam3r:
A couple of questions.
1. Is reflection really required here?
2. Is it possible to do more simple way for handling `[Readonly]Memory<byte>` as the most frequent case?

hrai:
> 
> 
> A couple of questions.
> 
>     1. Is reflection really required here?
> 
>     2. Is it possible to do more simple way for handling `[Readonly]Memory<byte>` as the most frequent case?

There's an extension method that we could have used but it's only available in .NET Core and .NET Standard. As shouldly targets .NET Framework, I don't think there's another way.

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.toenumerable?view=netcore-3.1

The code blocks were added by @jnm2 as it shows in https://github.com/shouldly/shouldly/issues/562

Perhaps he will have more info.

jnm2:
Shouldly used to build for .NET Framework 4.0 which meant it couldn't reference https://www.nuget.org/packages/System.Memory in its net40 compile and reflection seemed simpler than splitting the behavior. Since then, Shouldly has dropped support for net40. That means you can add an package reference rather than reflection.

hrai:
As `Memory<>` and `ReadOnlyMemory<>` are generic types, reflection seems to be the only way to cast to these types at runtime. When the equality method is invoked, there's no way to determine what the generic parameter's type is going to be.

This is what I found.
https://stackoverflow.com/questions/7676989/how-to-cast-a-generic-type-at-runtime-in-c-sharp

sungam3r:
I  talked about _explicit_ check for `[Readonly]Memory<byte>` and then cast to `IEnumerable` for the most frequent case. Otherwise fallback to reflection.

jnm2:
Ah, and I forgot I generics just now, quite right.

@sungam3r It would be indistinguishable to first check for IEnumerable, then fall back to using reflection to check for Memory/ReadOnlyMemory. Before we add additional checks on top of that, we need hard numbers on which scenarios it speeds up and which is slows down.

sungam3r:
I think that all scenarios will get a slowdown due to the fact that `IEnumerable` is used. All elements of value types are boxed to be used in `if (!equalityComparer.Equals(enumeratorX.Current, enumeratorY.Current))`.

jnm2:
@sungam3r With regard to that, this PR doesn't change anything. It's good to keep changes small and focused. Because of how much more complex it makes the code to use generic IEnumerable, I'd also prefer to see hard measurements of real-world projects demonstrating the problem so that we know we're doing something meaningful.

jnm2:
Also, if we were to go the direction of avoiding boxing, it would be both simpler and more general to first have a stage where an non-generic IEnumerable is obtained and then a second stage where generic IEnumerable is detected with a specialized iteration implementation that we invoke using MakeGenericMethod. That way not just `ReadOnlyMemory<byte>`  but any memory/enumerable of any value type gets the benefit. And that way, this PR doesn't need to be concerned about it one way or the other.

sungam3r:
I understand all this and do not insist.

jnm2:
@sungam3r I appreciate that. Thanks for your input. Would you be interested in starting a new issue and doing a benchmark to help us understand how big of an effect we're talking about, or do you want to wait and see if someone starts noticing some pain over it?

sungam3r:
No, I don’t think it matters much.

hrai:
It doesn't look like any more changes are expected in this pr. Can I get the approval then @sungam3r ?

src/Shouldly.Tests/ShouldBe/EnumerableType/EnumerableOfReadOnlyMemoryScenario.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            Verify.ShouldFail(() =>
+                    foo.ShouldBe(bar, "Some additional context"),
+
+errorWithSource:
+@"foo
+    should be
+System.ReadOnlyMemory<Byte>[2]
+    but was
+System.ReadOnlyMemory<Byte>[3]
```
jnm2: Normally the elements are listed in the error message. Could you see what it would take to enumerate here too?

src/Shouldly/Internals/EqualityComparer.cs
```diff
@@ -91,6 +92,45 @@ public bool Equals(T x, T y)
...
+                }
+            }
+
+            return enumerable != null;
+        }
+
+        private static IEnumerable ToEnumerable(object readOnlyMemory, Type elementType)
+        {
+            return (IEnumerable)Type.GetType("System.Runtime.InteropServices.MemoryMarshal, System.Memory")
+                ?.GetMethod("ToEnumerable", BindingFlags.Public | BindingFlags.Static | BindingFlags.DeclaredOnly)
```
jnm2: It's best to specify parameter types so that this doesn't crash if an overload is added.

src/Shouldly.Tests/ShouldBe/EnumerableType/EnumerableOfReadOnlyMemoryScenario.cs
hrai: The method that generates the error method takes in the original and expected objects and passes it to this method - https://github.com/shouldly/shouldly/blob/5fc845c32ccbf2b7e8a6c9186d3cd6382d9ae430/src/Shouldly/Internals/ShouldlyCoreExtensions.cs#L22

We can convert actual and expected `Memory` items into Enumerables in
1.  `ShouldBe()` and work with Enumerables
2.  this line while constructing `ExpectedActualShouldlyMessage` https://github.com/shouldly/shouldly/blob/5fc845c32ccbf2b7e8a6c9186d3cd6382d9ae430/src/Shouldly/Internals/ShouldlyCoreExtensions.cs#L39

src/Shouldly/Internals/EqualityComparer.cs
hrai: I wanted to use this overload of `GetMethod()` but it only support .NET Standard 2.0 and above - https://docs.microsoft.com/en-us/dotnet/api/system.type.getmethod?view=netframework-4.8#System_Type_GetMethod_System_String_System_Reflection_BindingFlags_System_Reflection_Binder_System_Type___System_Reflection_ParameterModifier___

Below is the modified code which fails to compile.
```
            return (IEnumerable)Type.GetType("System.Runtime.InteropServices.MemoryMarshal, System.Memory")
                ?.GetMethod("ToEnumerable", BindingFlags.Public | BindingFlags.Static | BindingFlags.DeclaredOnly, null, new[] { typeof(ReadOnlyMemory<T>) }, null)
                ?.MakeGenericMethod(elementType)
                .Invoke(null, new[] { readOnlyMemory });

```

hrai: image

src/Shouldly/Internals/EqualityComparer.cs
```diff
@@ -91,6 +92,45 @@ public bool Equals(T x, T y)
...
+            enumerable = obj as IEnumerable;
+
+            if (enumerable == null && obj != null)
+            {
+                var type = obj.GetType();
+                if (type.IsMemory(out var elementType))
+                {
+                    var readOnlyMemory = type.GetMethods(BindingFlags.Public | BindingFlags.Static | BindingFlags.DeclaredOnly)
+                        .SingleOrDefault(method =>
+                            method.Name == "op_Implicit"
```
sungam3r: Also you can check `method.IsSpecialName` for operators.

src/Shouldly/Internals/EqualityComparer.cs
jnm2: Ah, that's too bad. We could cross our fingers and hope that an overload of MemoryMarshal.ToEnumerable is never added, or we could use GetMethods+SingleOrDefault like you did above, or we could replace netstandard1.3 with netstandard2.0.

@JosephWoodward What are your thoughts?

src/Shouldly.Tests/ShouldBe/EnumerableType/EnumerableOfReadOnlyMemoryScenario.cs
jnm2: Thanks, that helps. It looks like those values are used here:

https://github.com/shouldly/shouldly/blob/5fc845c32ccbf2b7e8a6c9186d3cd6382d9ae430/src/Shouldly/MessageGenerators/ShouldBeMessageGenerator.cs#L20-L21

And the string is built here:

https://github.com/shouldly/shouldly/blob/5fc845c32ccbf2b7e8a6c9186d3cd6382d9ae430/src/Shouldly/Internals/StringHelpers.cs#L42-L51

I'm thinking we should obtain the IEnumerable in ToStringAwesomely for consistency because it sometimes uses the type full name and ToString of the original instance, and we wouldn't want to use the instance returned from MemoryMarshal.ToEnumerable for that.

hrai: Pushed a change

src/Shouldly/Internals/EqualityComparer.cs
hrai: GetMethods+SingleOrDefault sounds like a small enough change as opposed to upgrading to netstandard2.0

The library may stop supporting older .NET versions which may have unintended consequences.

src/Shouldly.Tests/Shouldly.Tests.csproj
```diff
@@ -1,28 +1,29 @@
...
-  </ItemGroup>
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net472' ">
-    <Reference Include="System.Runtime" />
-    <Reference Include="System.Threading.Tasks" />
-    <Reference Include="System" />
-    <Reference Include="Microsoft.CSharp" />
-  </ItemGroup>
-</Project>
\ No newline at end of file
+    <PropertyGroup>
```
jnm2: Any chance you could reformat back to the original (standard) 2-space csproj indentation, and squash it so the most recent line history for every line in the file doesn't point back to this PR?

src/Shouldly/Internals/StringHelpers.cs
```diff
@@ -38,12 +36,23 @@ internal static string ToStringAwesomely(this object value)
...
+            {
+                var readOnlyMemory = value.ToReadOnlyMemory(objectType, genericParameterType);
+                value = readOnlyMemory.ToEnumerable(genericParameterType);
+            }
+            else if (objectType.IsReadOnlyMemory(out genericParameterType))
+            {
+                value = value.ToEnumerable(genericParameterType);
+            }
 
             if (value is IEnumerable)
```
jnm2: What are your thoughts on using `TryGetEnumerable` from here?

SimonCropp:
thanks peoples

PR #619, opened 2020-06-03
force correct debug settings

### LINKED ISSUE #551: Extra Debug mode information added in Release mode.
I am facing issue with 3.0.2 Version of Shouldly. My Test Project is .net framework 461. I am running tests in Debug mode, Tests run fine but whenever I run the tests in Release mode, I am getting additional text added to test messages which fails my few tests as i use .Shouldbe Method. 

The information that is added is "Shouldly uses your source code to generate its great error messages, build your test project with full debug information to get better error messages"

Additional information: The issue comes wherever i am using

> Should.Throw<ShouldAssertException>(()=>Method that throws the exception)).Message.ShouldBe("My Error message")

SimonCropp:
fixes #551

### DIFF
--- a/src/Shouldly/Shouldly.csproj
+++ b/src/Shouldly/Shouldly.csproj
@@ -7,9 +7,9 @@
     <AssemblyOriginatorKeyFile>sn.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
-  
+
     <Description>Shouldly - Assertion framework for .NET. The way asserting *Should* be</Description>
-    <Authors>Jake Ginnivan, Joseph Woodward</Authors>    
+    <Authors>Jake Ginnivan, Joseph Woodward</Authors>
     <PackageTags>test;unit;testing;TDD;AAA;should;testunit;rspec;assert;assertion;framework</PackageTags>
     <PackageIcon>assets/logo_128x128.png</PackageIcon>
     <PackageProjectUrl>https://github.com/shouldly/shouldly</PackageProjectUrl>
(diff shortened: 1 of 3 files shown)


slang25:
I don't think we want to go down this route, full pdbs are a legacy format. We need to make it clear how portable pdbs can be loaded on older .NET Framework IMO.

slang25:
See here: https://github.com/shouldly/shouldly/pull/548

JosephWoodward:
@slang25 I'm inclined to agree. What's the state of #548?

SimonCropp:
i dont think #548 will fix things. have you smoke tested it? i did a temp PR with the test targeting portable and they still work https://github.com/shouldly/shouldly/pull/620. i think u would need a machine with only 461 installed to properly verify it

Either way i think we are conflating two things. https://github.com/shouldly/shouldly/issues/551 makes not mention of portable pdb. (yes they are problematic in some older runtimes). the problem with 551 is they did not have the debug info available to get the source information. this PR forces debug info to be available in all consuming projects without the need for them to set:

```
 <Optimize>false</Optimize>
    <DebugType>full</DebugType>
```

as the workaround describes: https://github.com/shouldly/shouldly/issues/551#issuecomment-453981071

i targeted `DebugType full` in this pr to support people running older runtimes that do not even have the concept of what a portable pdb is. happy to move to portable (although my preference is embedded) but if we did that i would recommend moving to the min supported "classic runtime"  to be 4.8

slang25:
Hey, I'll take another look at some point today. The thing with #551 was they changed the settings in their project, there's a couple of things to distinguish:
- Shouldly walks the stack to determine where the boundary between a shouldly assertion is used vs other shouldly goop, the stack walking needs to work, either it needs to deal with no symbols, or we need to force symbols (I think we're now good with the first, this has changed a lot over time). It also needs to be not optimized (or deal with optimized), as this squashes some of the stack frames at runtime.
- Shouldly will print a friendly message based on your consuming code, only if you have the relevant debug settings (optimizations shouldn't matter here, but that's another thing in the mix)

It's been so long since I looked at this, I may be conflating issues, and some may no longer be issues.

SimonCropp:
@JosephWoodward thoughts?

PR #634, opened 2020-07-06
Jnm2 nullable reference types

### LINKED ISSUE #579: Provide annotations for C# 8 Nullable Reference Types

SimonCropp:
replaces https://github.com/shouldly/shouldly/pull/581

Closes #579.


jnm2:
This was a squash merge 😢 I can't fetch these commits and look through them locally.

SimonCropp:
@jnm2 sorry about the hassle. you should be able to delete you master and re-pull this master. or delete and re-fork

jnm2:
It's okay. It's not that, it's that the commits in this PR don't exist in the upstream master branch (or the entire upstream repo). Only a squash commit exists:

image

PR #641, opened 2020-07-07
Redo nullable reference types fixes from original branch

jnm2:
Requested by @SimonCropp at https://github.com/shouldly/shouldly/pull/581#issuecomment-654060144

Please review a commit at a time. Most of this was cherry-picked from #581. I'm happy to find alternatives for any of the commits. There were significant problems with the structures that the three "Rewrite ..." commits address. We need to make the logic easier to follow. My commits very well may not have done this, but if not, we need to work toward something that does.

⚠ GitHub lists the commits in order of date rather than parent-child order. This means that the four cherry-picks appear together at the top instead of at the bottom of the list.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -5,4 +5,9 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="TunnelVisionLabs.ReferenceAssemblyAnnotator" Version="1.0.0-alpha.154" PrivateAssets="all" />
+    <PackageDownload Include="Microsoft.NETCore.App.Ref" Version="[3.1.0]" />
+  </ItemGroup>
+
 </Project>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -5,9 +5,4 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
-  <ItemGroup>
-    <PackageReference Include="TunnelVisionLabs.ReferenceAssemblyAnnotator" Version="1.0.0-alpha.154" PrivateAssets="all" />
-    <PackageDownload Include="Microsoft.NETCore.App.Ref" Version="[3.1.0]" />
-  </ItemGroup>
-
 </Project>
--- a/src/Shouldly/Configuration/FirstNonShouldlyMethodFinder.cs
+++ b/src/Shouldly/Configuration/FirstNonShouldlyMethodFinder.cs
@@ -1,6 +1,6 @@
 #if ShouldMatchApproved
+using System;
 using System.Diagnostics;
-using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Text.RegularExpressions;
@@ -20,20 +20,23 @@ namespace Shouldly.Configuration
 
         public TestMethodInfo GetTestMethodInfo(StackTrace stackTrace, int startAt = 0)
         {
-            var i = startAt;
-            StackFrame callingFrame;
-            do
+            for (var i = startAt; stackTrace.GetFrame(i) is { } frame; i++)
             {
-                callingFrame = stackTrace.GetFrame(i++);
-            } while (callingFrame.GetMethod().IsShouldlyMethod() || IsCompilerGenerated(callingFrame.GetMethod()));
+                if (frame.GetMethod() is { } method && !method.IsShouldlyMethod() && !IsCompilerGenerated(method))
+                {
+                    var callingFrame = stackTrace.GetFrame(i + Offset)
+                        ?? throw new InvalidOperationException("There is no stack frame at the specified offset from the first non-Shouldly stack frame.");
 
-            callingFrame = stackTrace.GetFrame(i + Offset - 1);
-            return new TestMethodInfo(callingFrame);
+                    return new TestMethodInfo(callingFrame);
+                }
+            }
+
+            throw new InvalidOperationException("Cannot find a non-Shouldly method in the stack trace.");
         }
 
         static bool IsCompilerGenerated(MethodBase method)
         {
-            return method.GetCustomAttributes(typeof(CompilerGeneratedAttribute), true).Any() || AnonMethod.IsMatch(method.Name);
+            return method.IsDefined(typeof(CompilerGeneratedAttribute), inherit: true) || AnonMethod.IsMatch(method.Name);
         }
     }
 }
--- a/src/Shouldly/Internals/EqualityComparer.cs
+++ b/src/Shouldly/Internals/EqualityComparer.cs
@@ -14,7 +14,6 @@ namespace Shouldly
     internal class EqualityComparer<T> : IEqualityComparer<T>
     {
         static readonly IEqualityComparer DefaultInnerComparer = new EqualityComparerAdapter<object>(new EqualityComparer<object>());
-        static readonly Type NullableType = typeof(Nullable<>);
 
         readonly Func<IEqualityComparer> _innerComparerFactory;
 
@@ -26,20 +25,11 @@ namespace Shouldly
 
         public bool Equals([AllowNull] T x, [AllowNull] T y)
         {
-            var type = typeof(T);
-
             if (ReferenceEquals(x, y))
                 return true;
 
-            // Null?
-            if (!type.IsValueType() || (type.IsGenericType() && type.GetGenericTypeDefinition().IsAssignableFrom(NullableType)))
-            {
-                if (Equals(x, null))
-                    return Equals(y, null);
-
-                if (Equals(y, null))
-                    return false;
-            }
+            if (x is null || y is null)
+                return false;
 
             if (Numerics.IsNumericType(x) &&
                 Numerics.IsNumericType(y))
@@ -49,8 +39,8 @@ namespace Shouldly
             }
 
             // Enumerable?
-            if (x!.TryGetEnumerable(out var enumerableX) &&
-                y!.TryGetEnumerable(out var enumerableY))
+            if (x.TryGetEnumerable(out var enumerableX) &&
+                y.TryGetEnumerable(out var enumerableY))
             {
                 var enumeratorX = enumerableX.GetEnumerator();
                 var enumeratorY = enumerableY.GetEnumerator();
--- a/src/Shouldly/Internals/EqualityComparer.cs
+++ b/src/Shouldly/Internals/EqualityComparer.cs
@@ -94,7 +94,6 @@ namespace Shouldly
             return object.Equals(x, y);
         }
 
-        public int GetHashCode([DisallowNull] T obj)
-            => throw new NotImplementedException();
+        public int GetHashCode(T obj) => throw new NotImplementedException();
     }
 }
--- a/src/Shouldly/Internals/EqualityComparer.cs
+++ b/src/Shouldly/Internals/EqualityComparer.cs
@@ -71,11 +71,11 @@ namespace Shouldly
 
             // Implements IEquatable<T>?
             if (x is IEquatable<T> equatable)
-                return equatable.Equals(y!);
+                return equatable.Equals(y);
 
             // Implements IComparable<T>?
             if (x is IComparable<T> comparableGeneric)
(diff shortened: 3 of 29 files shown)


src/Shouldly/Internals/SourceCodeTextGetter.cs
```diff
@@ -58,7 +58,7 @@ void ParseStackTrace(StackTrace? trace)
                 // walk up stack trace until we get to the calling method might have to be updated regularly as we find more
                 // scenarios. Alternately, it could be replaced with a more robust implementation.
                 while (currentFrame.GetMethod().DeclaringType == null ||
-                        (currentFrame.GetMethod().DeclaringType.FullName?.StartsWith("System.Dynamic") ?? false))
+                        (currentFrame.GetMethod().DeclaringType.FullName?.StartsWith("System.Dynamic", StringComparison.Ordinal) ?? false))
```
jnm2: In two separate instances, StartsWith/EndsWith with the default CurrentCulture comparison in NUnit caused such bad performance for Jon Skeet that he reported it. For NUnit I'm proposing adding Fxcop Analyzers, but I'd recommend at least doing a one-time sweep.

src/Directory.Build.props
```diff
@@ -5,4 +5,9 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <ItemGroup>
```
SimonCropp, line 8: why are these required?

src/Directory.Build.props
```diff
@@ -5,4 +5,9 @@
...
+    <PackageReference Include="TunnelVisionLabs.ReferenceAssemblyAnnotator" Version="1.0.0-alpha.154" PrivateAssets="all" />
+    <PackageDownload Include="Microsoft.NETCore.App.Ref" Version="[3.1.0]" />
+  </ItemGroup>
```
jnm2, line 11: This brings two significant benefits:

- It reduces the number of times you have to add nullability suppressions (`!` operator) because it clues the compiler in to e.g. the fact that `string.IsNullOrEmpty` is a null check.
- It catches places where the BCL is documented to sometimes return null which would cause Shouldly to throw NullReferenceException.

src/Shouldly/Configuration/FirstNonShouldlyMethodFinder.cs
```diff
@@ -20,20 +20,23 @@ public class FirstNonShouldlyMethodFinder : ITestMethodFinder
...
+                    var callingFrame = stackTrace.GetFrame(i + Offset)
+                        ?? throw new InvalidOperationException("There is no stack frame at the specified offset from the first non-Shouldly stack frame.");
 
-            callingFrame = stackTrace.GetFrame(i + Offset - 1);
-            return new TestMethodInfo(callingFrame);
+                    return new TestMethodInfo(callingFrame);
+                }
+            }
+
+            throw new InvalidOperationException("Cannot find a non-Shouldly method in the stack trace.");
```
jnm2, line 34: Recommend using the split view for these sections. If any of this is too exotic, I'm more than happy to rewrite.

src/Directory.Build.props
jnm2: This brings two significant benefits:

- It reduces the number of times you have to add nullability suppressions (`!` operator) because it clues the compiler in to e.g. the fact that `string.IsNullOrEmpty` is a null check.
- It catches places where the BCL is documented to sometimes return null which would cause Shouldly to throw NullReferenceException.

SimonCropp: ok cool. but given nuget ui doesnt yet recognize solutionlevel packages, do u think it would be better (for the short term) to make them project level to make it easier to update them?

src/Shouldly/Configuration/FirstNonShouldlyMethodFinder.cs
```diff
@@ -20,20 +20,23 @@ public class FirstNonShouldlyMethodFinder : ITestMethodFinder
...
         public TestMethodInfo GetTestMethodInfo(StackTrace stackTrace, int startAt = 0)
         {
-            var i = startAt;
-            StackFrame callingFrame;
-            do
+            for (var i = startAt; stackTrace.GetFrame(i) is { } frame; i++)
             {
-                callingFrame = stackTrace.GetFrame(i++);
-            } while (callingFrame.GetMethod().IsShouldlyMethod() || IsCompilerGenerated(callingFrame.GetMethod()));
+                if (frame.GetMethod() is { } method && !method.IsShouldlyMethod() && !IsCompilerGenerated(method))
```
SimonCropp, line 25: `frame.GetMethod() is { } method`  what does that syntax mean?

src/Shouldly/Configuration/FirstNonShouldlyMethodFinder.cs
SimonCropp: good point

src/Shouldly/Internals/SourceCodeTextGetter.cs
SimonCropp: thanks. i had no idea. is this still a problem in the current netcore?

src/Shouldly/Configuration/FirstNonShouldlyMethodFinder.cs
jnm2: It means if `frame.GetMethod()` is not null, put it in the `method` variable and evaluate true. Otherwise evaluate false.

It's a degenerate case of C# 8 property patterns. A non-degenerate case would actually have at least one property:

`if (frame.GetMethod() is { Name: "Foo" } method)`

jnm2: If you like, I'll rewrite as:

```cs
var method = frame.GetMethod();
if (method != null && !method.IsShouldlyMethod() && !IsCompilerGenerated(method))
```

SimonCropp: not at all. i am enjoying the education

src/Directory.Build.props
jnm2: I found some superfluous existing uses of `!` that can be removed now that the BCL is annotated. I'll push that soon.

jnm2: Two commits pushed actually.

src/Shouldly/Internals/EqualityComparer.cs
```diff
@@ -26,20 +25,11 @@ public EqualityComparer(IEqualityComparer? innerComparer = null)
...
                 return true;
 
-            // Null?
-            if (!type.IsValueType() || (type.IsGenericType() && type.GetGenericTypeDefinition().IsAssignableFrom(NullableType)))
-            {
-                if (Equals(x, null))
-                    return Equals(y, null);
-
-                if (Equals(y, null))
-                    return false;
```
jnm2, line 41: Object.Equals returns false if *either* parameter is null. Therefore these are equivalent to `x is null` and `y is null`.

`y is null` is false if `x is null` because of the ReferenceEquals check at the beginning.

The `if (!type.IsValueType() || (type.IsGenericType() && type.GetGenericTypeDefinition().IsAssignableFrom(NullableType)))` check is a very long way to check `default(T) is null`. Neither check is needed because if `default(T)` is not null, then `x` and `y` cannot be null and therefore the null checks are no-ops. (They are even elided during jit.)

src/Shouldly/Internals/SourceCodeTextGetter.cs
jnm2: I'm not sure, but I do know that the semantics issue is the same and there is still perf overhead associated with doing linguistic comparisons instead of binary.

jnm2:
@SimonCropp I'm comparing the diff between my original branch and this new branch. I see that there is a net difference that includes some missing `?`'s in the core assertions. Should I push more commits here or branch off for another PR so you can fix the master warnings sooner?

src/Shouldly/Configuration/FirstNonShouldlyMethodFinder.cs
jnm2: Oh, great! 😀

src/Directory.Build.props
jnm2: > ok cool. but given nuget ui doesnt yet recognize solutionlevel packages, do u think it would be better (for the short term) to make them project level to make it easier to update them?

Forgot to refresh, just saw this. I am used to updating those by copying and pasting the PackageReference tag from the NuGet page, but I do get what you mean. Sounds good. I'll push a commit for this next.

(There's something slightly different than Directory.Build.props around the corner, [Directory.Packages.props](https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions), but that's not shipped yet.)

jnm2: Done.

SimonCropp:
> Should I push more commits here

lets continue here. i will hold off any changes that might conflict until you are done

src/Shouldly/Internals/SourceCodeTextGetter.cs
jnm2: See https://github.com/nunit/nunit/pull/3583 for Jon's comment and links elsewhere

jnm2: The second one was just last month, perf hit was 5-6x

SimonCropp: perhaps i should re-purpose this code https://github.com/Fody/Caseless to make `StartsWith("System.Dynamic")` compile to  `StartsWith("System.Dynamic", StringComparison.Ordinal)` :)

jnm2:
@SimonCropp Stack_trace_is_trimmed_the_same_as_default_exception_stack_traces failed the PR validation build but it passes on my machine. Looks like master is failing, so I'll ignore for this PR if you don't think it's related.

src/Shouldly/Internals/SourceCodeTextGetter.cs
jnm2: That sounds cool, but there's already so much confusion around these APIs, idk about adding contextual subtleties...

SimonCropp:
yeah there are two tests that fail intermittently. they are on my list

SimonCropp:
and yes ignore them

jnm2:
Alright, I'm done! It was much less than I thought at first glance.

The ShouldlyApi test doesn't start properly (I worked around it by removing the preprocessor and running the net48 build):

```
  Message: 
    System.ComponentModel.Win32Exception : The specified executable is not a valid application for this OS platform.
  Stack Trace: 
    Process.StartWithCreateProcess(ProcessStartInfo startInfo)
    Process.Start()
    Process.Start(ProcessStartInfo startInfo)
    Process.Start(String fileName, String arguments)
    DiffTool.Open(String receivedPath, String approvedPath, Boolean approvedExists) line 53
    ShouldMatchApprovedTestExtensions.ShouldMatchApproved(String actual, Func`1 customMessage, Action`1 configureOptions) line 79
    ShouldMatchApprovedTestExtensions.ShouldMatchApproved(String actual, Action`1 configureOptions) line 25
    ApprovePublicApi.ShouldlyApi() line 16
```

First guess is that UseShellExecute defaults to false on .NET Core but true on .NET Framework.

SimonCropp:
awesome work. do you want me to squash?

jnm2:
I appreciate you asking! Squash merges make me sad. If the goal is clean commit history, most Git tools have a first-parent view that is the same. Squashing prevents future investigators from checking out or bisecting within the merged branch a year from now. (This has happened to me and it was quite limiting.) But most important is that commits tell a story. Their granularity and messages were cultivated with some effort on my part to document themselves. In my day job and OSS work, I'm often tracing things backwards. I know what it's like to come across a squash commit. Overall, squashing feels like a loss.

However, I know other people see commits differently or deal with messier situations. No worries either way. If allowed, I'd rather squash things down to an acceptable size myself before merging. I stop force-pushing when reviewing starts so that reviewers can keep their place more easily.

Thanks for working so hard getting these PRs in!

SimonCropp:
agreed. sometimes the commit history is valuable. sometimes it is a mess. i think this one is good enough to merge as is

PR #650, opened 2020-08-03
Modify FindAsyncMethodUsingAttribute to discover async test methods.

Kermittt:
This PR fixes `FindMethodUsingAttribute` so that it will correctly find the test method when the test method is async.

If you use `ShouldMatchApproved` in an async test method, after an await, the continuation that is created causes the stack trace to loose information about the original calling method.

e.g. This will fail with "Cannot find method in call stack with attribute NUnit.Framework.TestAttribute"
```
        [Test]
        public async Task ShouldMatchSnapshot_AsyncTestWithTestAttribute_ShouldMatchStringValueSnapshot()
        {
            await Task.Delay(1);

            const string stringValue = "String Value";
            stringValue.ShouldMatchApproved(c => c.LocateTestMethodUsingAttribute<TestAttribute>());
        }
```

While investigating this problem I realized that the `TestMethodInfo` class already appears to be aware of this issue and includes code to resolve the async methods, however this has not been implemented in `FindMethodUsingAttribute`.

To fix this, I've copied a number of classes/methods from `TestMethodInfo` and applied them in `FindMethodUsingAttribute`.  Note that I've copied `OriginalMethodInfo`, `GetOriginalMethodInfoForStateMachineMethod` and `ContainsAttribute`, and these are all unchanged, so it may be worth refactoring them to a shared location later.

If anyone wishes to make use of this before the PR is merged, you can simply define a new `FindMethodUsingAttribute` with a defferent name (e.g. `FindAsyncMethodUsingAttribute`, and then configure `ShouldMatchApproved` to use it like so...
`stringValue.ShouldMatchApproved(c => c.Configure(smc => smc.TestMethodFinder = new FindAsyncMethodUsingAttribute<TestAttribute>()));`

### DIFF
--- a/src/Shouldly.Tests/ShouldMatchApproved/ShouldMatchApprovedScenarios.CanFindTestAttributeInAsync.approved.txt
+++ b/src/Shouldly.Tests/ShouldMatchApproved/ShouldMatchApprovedScenarios.CanFindTestAttributeInAsync.approved.txt
@@ -1,0 +1,1 @@
+testAttributes
--- a/src/Shouldly.Tests/ShouldMatchApproved/ShouldMatchApprovedScenarios.cs
+++ b/src/Shouldly.Tests/ShouldMatchApproved/ShouldMatchApprovedScenarios.cs
@@ -168,6 +168,14 @@ In the meantime use 'ShouldlyConfiguration.DiffTools.RegisterDiffTool()' to add 
         }
 
         [Fact]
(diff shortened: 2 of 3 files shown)


SimonCropp:
thanks. can you add that failing test to the PR?

Kermittt:
> thanks. can you add that failing test to the PR?

No problem.  Have done so.

SimonCropp:
thanks 👍

PR #676, opened 2020-09-23
Restore the ability to use Should(Not)BeNull on System.Nullable<T> values

### LINKED ISSUE #674: int?.ShouldBeNull issue
It looks like in the latest 0003 release we've broken the following scenarios when using `Nullable<T>` due to the [class constraint](https://github.com/shouldly/shouldly/issues/584):

```csharp
int? age = null;
age.ShouldBeNull();

int? age = 5;
age.ShouldNotBeNull();
```

```
[CS0452] The type 'int?' must be a reference type in order to use it as parameter 'T' in the generic type or method 'ShouldBeNullExtensions.ShouldNotBeNull<T>(T?, string?)'
```

jnm2:
Fixes #674. Adds tests that would have brought this to our attention in #626.

I think it was a recent backported compiler change that allows this kind of overload resolution, though maybe this has been around longer than I'm thinking too:

```cs
public static void ShouldBeNull<T>(this T? x) where T : class { }
public static void ShouldBeNull<T>(this T? x) where T : struct { }
```

```cs
((string?)null).ShouldBeNull(); // No error!
((int?)null).ShouldBeNull(); // No error!
4.ShouldBeNull(); // Error!
```

### DIFF
--- a/src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
+++ b/src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
@@ -402,16 +402,14 @@ namespace Shouldly
     {
         [System.Obsolete("Func based customMessage overloads have been removed. Pass in a string for the cu" +
             "somMessage.", true)]
-        public static void ShouldBeNull<T>(this T? actual, System.Func<string?>? customMessage)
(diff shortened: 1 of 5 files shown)


SimonCropp (approved):

JosephWoodward:
@jnm2 Great, thanks for this!

PR #680, opened 2020-10-01
fix: Add link to Shoudly

NauhcNoohc:
Hi, I have added link to **Shoudly** `https://docs.shouldly.io/` in _README.md_


SimonCropp:
i dont think it is correct to link every instance of "shouldly" to the docs site. especially when there is already one on the top right of the github page

SimonCropp:
if you want to contribute, there are plenty of issues that we could use the help in triaging

NauhcNoohc:
Alright. 
Side topic. I saw there is a duplicate `README.md` in this project which can be found in
[https://github.com/shouldly/shouldly/blob/master/documentation/README.md](url) and
[https://github.com/shouldly/shouldly/blob/master/README.md](url)

Also, both content are not identical. Would you mind i update this?

SimonCropp:
`documentation/README.md ` is the content that will be rendered into read the docs. while they may end up being similar, the work hasnt been done to sort out how similar. so there is not any current value to make them exactly the same

Issue #584, opened 2019-10-16
Disable accidental use of ShouldBeNullExtensions on structs, add "where T : class" on methods


r-pankevicius:
Assume code today, when people are crazy about use of var-s:

```csharp
var result = ClassUnderTest.MyMethod();
result.ShouldNotBeNull();
```

If type of result is int, then this check will succeed, for example 5 != null evaluates to true.
But this check is useless first of all. Maybe people used different type yesterday, today refactored it to use int, but compiler did not tell, hey, it's false sense of success.

I suggest adding generic method specification requiring value under test to be nullable, e.g. class, not struct.

Below is sample code to show the idea: ShouldNotBeNull_V2 is better than ShouldNotBeNull_V1 because compiler will tell you're doing something wrong.

```csharp
static class MyExtensions
{
	public static void ShouldNotBeNull_V1<T>(this T actual)
	{
		if (actual == null)
			throw new InvalidOperationException();
	}

	public static void ShouldNotBeNull_V2<T>(this T actual) where T : class
	{
		if (actual == null)
			throw new InvalidOperationException();
	}
}

class UserCode
{
	public void Test()
	{
		int x = 5;

		5.ShouldNotBeNull_V1(); // possible to write it
		5.ShouldNotBeNull_V2(); // you get compiler error
	}
}
```

sungam3r:
Hmm, I think that there already was where constraint ).

r-pankevicius:
I was looking at
https://github.com/shouldly/shouldly/blob/master/src/Shouldly/ShouldlyExtensionMethods/ShouldBeNullExtensions.cs
today

r-pankevicius:
...and I've set up latest stable version of Shouldly: Assembly Shouldly, Version=3.0.2.0, Culture=neutral, PublicKeyToken=6042cbcb05cbc941, and was able to write:

```csharp
[Fact]
public void ICanHaz()
{
	5.ShouldNotBeNull();
	5.ShouldBeNull();
}
```
Well, compiler did not cry for Argentina.

$>ping @sungam3r

sungam3r:
Well, in small steps we will sooner or later come to the ideal.

r-pankevicius:
@sungam3r  Send a letter to Anders Hejlsberg, because C# compiler should catch this issue before us.

jnm2:
[...]

PR #684, opened 2020-10-04
Make ShouldNotBeNull chainable and unwrap System.Nullable

### LINKED ISSUE #559: ShouldHaveValue() for chainable assertions on nullable values

jnm2:
Closes #559 (supersedes the design)

I noticed that the new ShouldHaveValue() method would be completely identical to the struct variant of ShouldNotBeNull that we already ended up with. Instead of adding a new method, I augmented the existing ones. Making ShouldNotBeNull chainable for reference types was already on my wish list.

```cs
int unwrapped = ((int?)42).ShouldNotBeNull();
string worksWithNRTsToo = ((string?)"42").ShouldNotBeNull();
```

### DIFF
--- a/src/Shouldly/ShouldlyExtensionMethods/ShouldBeNullExtensions.cs
+++ b/src/Shouldly/ShouldlyExtensionMethods/ShouldBeNullExtensions.cs
@@ -26,19 +26,17 @@ namespace Shouldly
         }
 
         [ContractAnnotation("actual:null => halt")]
-        public static void ShouldNotBeNull<T>([NotNull] this T? actual, string? customMessage = null)
+        public static T ShouldNotBeNull<T>([NotNull] this T? actual, string? customMessage = null)
             where T : class
         {
-            if (actual == null)
-                throw new ShouldAssertException(new ExpectedShouldlyMessage(actual, customMessage).ToString());
+            return actual ?? throw new ShouldAssertException(new ExpectedShouldlyMessage(actual, customMessage).ToString());
         }
 
         [ContractAnnotation("actual:null => halt")]
-        public static void ShouldNotBeNull<T>([NotNull] this T? actual, string? customMessage = null)
+        public static T ShouldNotBeNull<T>([NotNull] this T? actual, string? customMessage = null)
             where T : struct
         {
-            if (actual == null)
-                throw new ShouldAssertException(new ExpectedShouldlyMessage(actual, customMessage).ToString());
+            return actual ?? throw new ShouldAssertException(new ExpectedShouldlyMessage(actual, customMessage).ToString());
         }
     }
 }
--- a/documentation/documentation/equality/null.md
+++ b/documentation/documentation/equality/null.md
@@ -1,7 +1,8 @@
 # BeNull
 
-`ShouldBeNull` and `ShouldNotBeNull` allow you to check whether or not a type's reference is null.
+`ShouldBeNull` and `ShouldNotBeNull` allow you to check whether a value is null.
 
+`ShouldNotBeNull` returns the non-null value if it succeeds so that further assertions can be chained. When used with a reference type, the returned value is the same reference annotated as non-null. Equivalently, when used on a `System.Nullable<T>` expression, the returned value is the unwrapped `T` value.
 
 ## ShouldBeNull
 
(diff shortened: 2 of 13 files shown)


jnm2:
@JosephWoodward @SimonCropp @slang25 The master branch is failing. Should we consider protecting it so that it can't be updated except via PRs that pass CI?

SimonCropp:
@jnm2 sorry the broken build is my fault. will look into it now

> Should we consider protecting it so that it can't be updated except via PRs that pass CI?

I am on the fence. but happy to go with the consensus

SimonCropp (commented):
can some docs be added for this change?

JosephWoodward:
> @JosephWoodward @SimonCropp @slang25 The master branch is failing. Should we consider protecting it so that it can't be updated except via PRs that pass CI?

I'm not so fussed personally. Whilst we should try our best not to break master, there's not enough change going through it which would mean it's broken for long.

src/Shouldly/ShouldlyExtensionMethods/ShouldBeNullExtensions.cs
```diff
@@ -26,19 +26,17 @@ public static void ShouldBeNull<T>([MaybeNull] this T? actual, string? customMes
         }
 
         [ContractAnnotation("actual:null => halt")]
-        public static void ShouldNotBeNull<T>([NotNull] this T? actual, string? customMessage = null)
+        public static T ShouldNotBeNull<T>([NotNull] this T? actual, string? customMessage = null)
```
RodgerLeblanc, line 29: Don't you want to return `T?` instead of `T`, even if it has the `NotNullAttribute` and that it will throw if the value is `null`?

It seems weird that a fluent method is not returning the same type as its parameter, meaning that the chained assertion will be performed on a different type.

For example, this won't compile : 
`((int?)0).ShouldNotBeNull().ShouldNotBeNull();`

jnm2: Thanks for the question! No, the motivation in #559 was specifically to unwrap nullable values. I have a lot of code asserts that a struct is not null and then asserts things about several deeper properties. Having to repeat `.Value` is onerous. There are code examples in #559 that show examples.

(This is also why we petitioned for C# 8's coalescing assignments to unwrap nullable values. The original design did not; it made you do `(nullableFoo ??= GetFoo()).Value`. Having to say `.Value` on a thing that is never null is noisy and frustrating. Thankfully, they listened and changed it before C# 8 shipped!)

I see it as a feature that `((int?)0).ShouldNotBeNull().ShouldNotBeNull();` doesn't compile, in line with https://github.com/shouldly/shouldly/issues/584, and I like the parallel with nullable reference types.

jnm2:
@SimonCropp Happy to, what kind of docs were you thinking? Adding something to DocumentationExamples, or something else?

SimonCropp:
>Adding something to DocumentationExamples

yep, and including those snippets into the markdown

jnm2:
@SimonCropp What do you do to generate the anchor IDs in the markdown?

src/Shouldly/ShouldlyExtensionMethods/ShouldBeNullExtensions.cs
jnm2: Another longstanding example of an assertion method that returns a different type is ShouldHaveSingleItem.

SimonCropp:
they will be updated on build https://github.com/SimonCropp/MarkdownSnippets

jnm2:
@SimonCropp I tried a few things based on https://github.com/SimonCropp/MarkdownSnippets, but `dotnet build src` and `dotnet test src` don't seem to be updating it. What am I missing?

jnm2:
Looks like the master branch is broken again due to merging https://github.com/shouldly/shouldly/pull/681.

SimonCropp:
>  I tried a few things based on https://github.com/SimonCropp/MarkdownSnippets, but dotnet build src and dotnet test src don't seem to be updating it. What am I missing?

Not sure. i think just merge this, and i will debug the doco

And i fixed the build

SimonCropp:
@jnm2 is this ready to be merged?

jnm2:
Yes, and it needs the doc tooling followup after merge that you mentioned.

PR #485, opened 2018-02-21
Equality comparison

whortleberrybearer:
Something that I started many months ago but never got round to putting it forward.  This adds the functionality detailed in #453.  The extent of the changes was mainly just adding overloads to the existing Should methods.

PR #687, opened 2020-10-10
Temp Equality comparison (#485)

### LINKED ISSUE #453: Addition of IEqualityComparer to some methods to allow custom comparisons
Would it be useful / suitable to allow an `IEqualityComparer<T>` instance to be passed into some methods to allow the equality checks to be customised?  E.g.

`public static void ShouldBe<T>(this T actual, T expected, IEqualityComparer<T> comparer) 
        {
            …
        }`

It’s similar to what has already been put forward in #364.

I have had a situation recently where I needed to compare 2 object, but overriding the `Equals` method was not suitable and I didn’t want to have to write many `x.Prop1.ShouldBe(y.Prop1); x.Prop2.ShouldBe(y.Prop2);` as it would have been a real maintenance pain.  I ended up with a comparison object that and just asserting that was true, e.g. `comparer.Equals(x, y).ShouldBeTrue()` which works, but lost the context in the failure message.  Allowing a comparer object to be passed into the ShouldBe method would allow the context to be retained, e.g. `x.ShouldBe(y, comparer)`.

I have noticed internally, `Is.Equal` already takes an `IEqualityComparer<T>` parameter, so much of the internal functionality is already in place, it would just need some new extension methods to allow the comparer to be passed it.

I’m thinking it would only be suitable in a few places, such as the generic `T` and `IEnumerable<T>` `ShouldBe` and `ShouldNotBe` methods.

Any thoughts?

SimonCropp:
replaces #485
fixes #453

### DIFF
--- a/src/Shouldly/ShouldlyExtensionMethods/ShouldBe/GenericShouldBeTestExtensions.cs
+++ b/src/Shouldly/ShouldlyExtensionMethods/ShouldBe/GenericShouldBeTestExtensions.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
@@ -23,13 +22,9 @@ namespace Shouldly
         }
         public static void ShouldBe<T>(this T actual, T expected, IEqualityComparer<T> comparer)
         {
-            ShouldBe(actual, expected, comparer, () => null);
+            ShouldBe(actual, expected, comparer, null);
         }
-        public static void ShouldBe<T>(this T actual, T expected, IEqualityComparer<T> comparer, string customMessage)
-        {
-            ShouldBe(actual, expected, comparer, () => customMessage);
-        }
-        public static void ShouldBe<T>(this T actual, T expected, IEqualityComparer<T> comparer, [InstantHandle] Func<string> customMessage)
+        public static void ShouldBe<T>(this T actual, T expected, IEqualityComparer<T> comparer, string? customMessage)
         {
             actual.AssertAwesomely(v => Is.Equal(v, expected, comparer), actual, expected, customMessage);
         }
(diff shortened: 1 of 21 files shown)


src/Shouldly/ShouldlyExtensionMethods/ShouldBe/GenericShouldBeTestExtensions.cs
```diff
@@ -12,20 +12,38 @@ public static partial class ShouldBeTestExtensions
...
         {
             if (ShouldlyConfiguration.CompareAsObjectTypes.Contains(typeof(T).FullName!) || typeof(T) == typeof(string))
                 actual!.AssertAwesomely(v => Is.Equal(v, expected, new ObjectEqualityComparer<T>()), actual, expected, customMessage);
             else
                 actual!.AssertAwesomely(v => Is.Equal(v, expected), actual, expected, customMessage);
         }
+        public static void ShouldBe<T>(this T actual, T expected, IEqualityComparer<T> comparer)
+        {
+            ShouldBe(actual, expected, comparer, null);
+        }
```
jnm2, line 26: All overloads added in this PR are missing the nullability annotations that are present on the methods they overload.

src/Shouldly/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
```diff
@@ -154,6 +199,22 @@ private static List<T> GetDuplicates<T>(IEnumerable<T> items)
...
+        // Could possible remove the above implementation and pass an ObjectComparer like the generic ShouldBe methods do.
+        // Also added generics instead of object as this is probably how it should have been implemented.
+        private static List<T> GetDuplicates<T>(IEnumerable<T> items, IEqualityComparer<T> comparer)
+        {
+            var list = new List<T>();
+            var duplicates = new List<T>();
+
+            foreach (var o1 in items)
+            {
+                duplicates.AddRange(list.Where(o2 => o1 != null && comparer.Equals(o1, o2)));
```
jnm2, line 211: This means `null` is never detected as a duplicate.

jnm2: Also, NUnit has struggled with performance issues by not sticking with hash lookups for these comparisons. I'd recommend using HashSet or Dictionary to make use of `IEqualityComparer<T>.GetHashCode`.

src/Shouldly/ShouldlyExtensionMethods/ShouldBe/GenericShouldBeTestExtensions.cs
jnm2: https://github.com/shouldly/shouldly/pull/700

src/Shouldly/ShouldlyExtensionMethods/ShouldBeEnumerableTestExtensions.cs
jnm2: https://github.com/shouldly/shouldly/pull/701

PR #692, opened 2020-10-11
Remove redundant dependency to System.Dynamic.Runtime.

teo-tsirpanis:
It is not needed for .NET Standard 2.0.

I would especially appreciate if this PR was marked with a `hacktoberfest-accepted` label.


SimonCropp:
@teo-tsirpanis thanks

PR #689, opened 2020-10-11
Simplify overloads using optional params

SimonCropp:

PR #699, opened 2020-10-16
Bring back obsolete overload that was deleted

jnm2:
The reason that the obsoleted overload existed prior to #689 was so that code that already called that overload would have a nicer error message than the C# compiler saying you used the wrong parameter types after you upgrade. That reason applies just as much after #689 as it did before it.


SimonCropp (approved):

jnm2:
Thanks @SimonCropp!

PR #702, opened 2020-10-16
Only mention each duplicate once

jnm2:
It sounded off to me to say "1, 1" was duplicated; the thing that was duplicated was what the duplicates all have in common. In this case "1"-ness.

@SimonCropp @JosephWoodward @slang25 This is more of a design change. Does anyone have a concern about it?


SimonCropp:
I am happy with this.

jnm2:
Rebased now that the base PR merged and unmarked as draft.

PR #730, opened 2020-12-14
move from targets to props

### LINKED ISSUE #723: Project referencing Shouldly NuGet package is rebuilt every time, because no .pdb file is generated
I created a trival C# project referencing Shouldly, like this:

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <OutputType>Library</OutputType>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.8.0" />
    <PackageReference Include="NUnit" Version="3.12.0" />
    <PackageReference Include="NUnit3TestAdapter" Version="3.17.0" />
    <PackageReference Include="Shouldly" Version="4.0.1" />
  </ItemGroup>

</Project>
```

`dotnet build` and Visual Studio build rebuild this project every time, even though no changes have been made. VS logs this:

> FastUpToDate: Output 'ShoudlyTest\bin\Debug\netcoreapp3.1\ShoudlyTest.pdb' does not exist, not up to date. (ShoudlyTest)

Indeed, that .pdb file is not generated by the build, even though the default settings are to generate a .pdb file. It still happens even if I try to explicitly set the project to generate a .pdb file:

```
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
    <DebugType>full</DebugType>
    <DebugSymbols>true</DebugSymbols>
  </PropertyGroup>
```

I see that build/Shouldly.targets in the NuGet package sets `<DebugType>embedded</DebugType>`, which is probably what causes this.

Even without the rebuild issue, Shouldly should not be changing the debug type settings for my project - I might need that .pdb file for other reasons.

SimonCropp:
fixes #723


drewnoakes:
How does this fix the issue? If the problem is related to when the properties are defined, then the `<Import>` element needs to be moved. The `.props` and `.targets` extensions are just a convention and don't influence anything at runtime.

drewnoakes:
Oh, never mind. These files are included in the package.

It might be worth moving the `<Import>` elements to the tops of the project files too, though.

SimonCropp:
@drewnoakes 

> It might be worth moving the <Import> elements to the tops of the project files too, though.

can u expand on why?

Also why would changing from targets to props fix the rebuild issue?

drewnoakes:
> why would changing from targets to props fix the rebuild issue

I didn't look into why so cannot say offhand. The props file would be evaluated _before_ any properties/items defined in the consumer's project, whereas the targets file would be evaluated _after_.

If that ordering is important to consumers of the package, it might be important in their usage in these projects too.

drewnoakes:
I stumbled upon this again and can provide some more context this time around.

Shouldly is setting `DebugType` to `embedded`. It used to do this in the `.targets` file, which would occur _after_ other properties were derived from the prior value, but before the compiler sampled the value.

Concretely:

1. MSBuild uses `DebugType` to determine whether debug symbols are going to be produced in [`Microsoft.Common.CurrentVersion.targets`](https://github.com/dotnet/msbuild/blob/c88325c7859bcd282ff1fae14ed95936e051e1c3/src/Tasks/Microsoft.Common.CurrentVersion.targets#L174-L182)
2. Shouldly's targets file was then changing the value of `DebugType`, which made it out-of-sync with `_DebugSymbolsProduced`
3. VS's design-time build (which powers its up-to-date check) sees `_DebugSymbolsProduced` is `true` in [`CollectUpToDateCheckBuiltDesignTime`](https://github.com/dotnet/project-system/blob/02f7d636d5bf224e1f10f214b7e62b544a4ea402/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets#L420-L422) target, and believes there should be a PDB file in the intermediate directory.
4. The compiler sees `DebugType` as `embedded` so no PDB file is created.
5. VS eternally sees the PDB file as missing and believes that the build should create it, so will always schedule a build.

PR #732, opened 2020-12-15
fix some nullability issues

SimonCropp:


src/Shouldly/Configuration/FindMethodUsingAttribute.cs
```diff
@@ -11,6 +11,10 @@ public TestMethodInfo GetTestMethodInfo(StackTrace stackTrace, int startAt = 0)
         {
             foreach (var frame in stackTrace.GetFrames().Skip(startAt))
             {
+                if (frame == null)
+                {
+                    continue;
+                }
```
jnm2, line 17: Because I moved to .NET 5 nullability annotations, `GetFrames()` is now properly annotated so that `frame` is known at compile time to never be null.

src/Shouldly/Internals/EqualityComparerAdapter.cs
```diff
@@ -19,7 +19,7 @@ public EqualityComparerAdapter(IEqualityComparer<T> innerComparer)
 
         public new bool Equals(object? x, object? y)
         {
-            return _innerComparer.Equals((T)x, (T)y);
+            return _innerComparer.Equals((T)x!, (T)y!);
```
jnm2, line 22: It would be preferable to instead stop casting to non-nullable T: `(T?)x`

src/Shouldly/Configuration/FindMethodUsingAttribute.cs
jnm2: Same thing in another file.

src/Shouldly/ShouldlyExtensionMethods/ShouldBeTypeTestExtensions.cs
```diff
@@ -9,7 +9,7 @@ public static partial class ShouldBeTestExtensions
         public static T? ShouldBeAssignableTo<T>(this object? actual, string? customMessage = null)
         {
             ShouldBeAssignableTo(actual, typeof(T), customMessage);
-            return (T)actual;
+            return (T?)actual;
```
jnm2, line 12: Good catch. Was there a warning for this? If not, the called ShouldBeAssignableTo overload is improperly annotated. If so, I wonder why I didn't see it.

src/Shouldly/Configuration/FindMethodUsingAttribute.cs
jnm2: This is what you look for:
image

jnm2: Also notice
image

jnm2: Resolved in https://github.com/shouldly/shouldly/pull/740

src/Shouldly/Internals/EqualityComparerAdapter.cs
jnm2: Resolved in https://github.com/shouldly/shouldly/pull/740

PR #757, opened 2021-01-22
Improve fail message of DateTime comparisons

### LINKED ISSUE #421: Improve fail message of DateTime comparisons
Currently, when comparing two datetimes, I'm getting this error:

```
Should.Core.Exceptions.GreaterThanOrEqualException: Assert.GreaterThanOrEqual() Failure:
	Expected: 2/8/2017 2:22:01 PM >= 2/8/2017 2:22:01 PM
	but it was not
```

After some debugging I found out that the problem was the ms part of the objects that differ.

I think it would be useful to add the ms part after the second. Something like:

```
Should.Core.Exceptions.GreaterThanOrEqualException: Assert.GreaterThanOrEqual() Failure:
	Expected: 2/8/2017 2:22:01.200 PM >= 2/8/2017 2:22:01.201 PM
	but it was not
```

Also, according to [DateTime.Equals Method](https://msdn.microsoft.com/en-us/library/wk5t802s(v=vs.110).aspx), it compares the datetimes' ticks, so even if the above is added, it wont help much when they have the same ms but differ in the ticks.

anthonyflores-github:
Fixes #421

### DIFF
--- a/src/Shouldly.Tests/ShouldBeGreaterOrEqualTo/StringScenario.cs
+++ b/src/Shouldly.Tests/ShouldBeGreaterOrEqualTo/StringScenario.cs
@@ -37,5 +37,13 @@ Additional Info:
         {
             "a".ShouldBeGreaterThanOrEqualTo("a");
         }
+
+        [Fact]
+        public void ShouldFailDate()
+        {
+            var oldDate = Convert.ToDateTime("22/1/2020 4:44:01.210 PM");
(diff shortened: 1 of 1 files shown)


jnm2:
There was a mixup and this had been fixed in https://github.com/shouldly/shouldly/pull/372 prior to the issue even being filed. We really appreciate your help on this though!

PR #758, opened 2021-01-24
Added ignore to ordering of enumerable in ShouldBeEquivalentTo

### LINKED ISSUE #679: How to check if two enumerations have the same values?
Hi,
I want to check if two collections with GUID have all the same items, but the order is not important. for me
`IEnumerable<Guid> list1;
IEnumerable<Guid> list2;`

I can use:
;`list1.ShouldBe(list2);`
but here the order is important.

Is there any possibility to do it nicely in Shouldly?

jamesmcroft:
## Fixes #679

A change has been made to the check of enumerable objects in the `ShouldBeEquivalentTo` method to ensure that the order is not taken into account as this is covered by the `ShouldBe` method also. 

This allows for a check on two collections that have the same items but in differing orders. Tests have been added for the scenario of the same but in a different order. Tests have also been updated for the changed error messaging and existing tests have been re-run to ensure they are still working as intended.

### DIFF
--- a/src/Shouldly.Tests/ShouldBeEquivalentTo/EnumerableScenario.cs
+++ b/src/Shouldly.Tests/ShouldBeEquivalentTo/EnumerableScenario.cs
@@ -77,18 +77,19 @@ Additional Info:
         public void ShouldFailWhenListsDoNotMatch()
         {
             var subject = new[] { 1, 2, 6, 4, 5 };
+
             Verify.ShouldFail(() =>
 subject.ShouldBeEquivalentTo(new[] { 1, 2, 3, 4, 5 }, "Some additional context"),
 
 errorWithSource:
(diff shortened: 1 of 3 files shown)


jnm2 (changes requested):
Hi, thanks for your help! Since `ShouldBe` on enumerables cares about order by default but stops caring when you call it with `ignoreOrder: true`, my default expectation would be that `ShouldBeEquivalentTo` would also have the same default and the same `ignoreOrder` parameter.

I'm also concerned about changing the behavior of something that has already shipped. We'd probably want to wait for a major version to make a change like this, and the release notes would need a glaring banner to tell people what the change is since their test code would silently change in meaning. From experience, people are often upset at this kind of change even with best effort at loud messaging.

PR #745, opened 2021-01-10
Enumerable traversal test

jswolf19:
This adds failing tests for issue #735 as per request by @SimonCropp.

`TestEnumerableIsNotRetraversable` is to illustrate how the test enumerable works.
This emulates the behavior a thin layer over a data stream (e.g. a csv file parser) might have.

The other four tests illustrate the expected behavior.
- ActualEnumerableShouldOnlyBeTraversedOnce
- ActualEnumerableShouldOnlyBeTraversedOnceWhenIgnoringOrder
- ExpectedEnumerableShouldOnlyBeTraversedOnce
- ExpectedEnumerableShouldOnlyBeTraversedOnceWhenIgnoringOrder

PR #759, opened 2021-01-26
Enumerate once

### LINKED ISSUE #735: `ShouldBe` on Enumerables assumes that an enumerable is traversable multiple times
When `ShouldBe` on an enumerable fails, it reenters the enumerable in order to make its error message. It is my understanding, however, that enumerables are not guaranteed to be retraversable, and a test on such an enumerable results in incorrect error reporting.

My use case is with testing a CSV parser that, when calling `ReadLine()` returns an enumerable dependent on an underlying `TextReader`, so that multiple traversals of the Enumerable result in different lines. I want to avoid extra buffering and requiring that the `TextReader` be seekable.

This can be worked around by using `ToList()`.

jnm2:
Closes #735, fixes failing tests in the master branch

Follows up https://github.com/shouldly/shouldly/pull/745#issuecomment-767323061

### DIFF
--- a/src/Shouldly.Tests/ShouldBe/EnumerableType/NonRetraversableEnumerable.cs
+++ b/src/Shouldly.Tests/ShouldBe/EnumerableType/NonRetraversableEnumerable.cs
@@ -47,7 +47,7 @@ namespace Shouldly.Tests.ShouldBe.EnumerableType
             var fooEnum = CreateTestEnumerable();
 
             Verify.ShouldFail(
-                () => fooEnum.ShouldBe(new[] { 2, 3, 4 }, true),
+                () => fooEnum.ShouldBe(new[] { 2, 3, 4 }, ignoreOrder: true),
                 errorWithSource:
 @"fooEnum
     should be (ignoring order)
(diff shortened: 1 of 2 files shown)


src/Shouldly.Tests/ShouldBe/EnumerableType/NonRetraversableEnumerable.cs
```diff
@@ -10,22 +10,20 @@ public class NonRetraversableEnumerable
...
 
         [Fact]
         public void ActualEnumerableShouldOnlyBeTraversedOnce()
         {
-            var fooEnum = CreateTestEnumerable();
-
-            Verify.ShouldFail(() =>
-                fooEnum.ShouldBe(new [] {3, 2, 1 }),
+            Verify.ShouldFail(
+                () => CreateTestEnumerable().ShouldBe(new[] { 3, 2, 1 }),
```
jnm2, line 24: If we wanted, we could make it more likely for contributors to avoid the pitfall of keeping mutable state outside the `Verify.ShouldFail` call by making all the code they are extrapolating from use `static () =>`.

SimonCropp:
@jnm2 this broke some other tests. do u have time to look at it?

SimonCropp:
https://ci.appveyor.com/project/shouldly/shouldly/builds/37564343/job/coiv59rfg35yiapn

jswolf19:
@SimonCropp @jnm2 It looks like [entry points I didn't add tests for](https://github.com/shouldly/shouldly/pull/745#issuecomment-762969754) generate the actual part differently... Or perhaps if the enumerable result is empty, the type info is added on? I can come up with some tests to suss out this behavior, if that would be helpful.

SimonCropp:
@jswolf19 

>  I can come up with some tests to suss out this behavior, if that would be helpful.

i think that would help

jswolf19:
@SimonCropp I'll try to get it done this weekend, but it might be next weekend ^^;

Issue #749, opened 2021-01-11
Truncate output of ShouldMatchApproved


loop-evgeny:
When ShouldMatchApproved fails it outputs

> should match approved with options: Ignoring line endings
> ... (whole approved file here)
>     difference
> Showing some of the 829476 differences

My "approved" file is ~900 KB and such an amount of text is not handled well by tools like Visual Studio and TeamCity. Those tools *should* of course handle it better, but Shouldly can easily work around it by truncating the output to some reasonable length - nobody is going to read 900 KB anyway.

SimonCropp:
what size would u truncate to ?

loop-evgeny:
It would take some experimentation to see what VS and TeamCity handle but off the top of my head, I'd say ~ 50-100 KB. Seems unlikely that a person would manually inspect more text that to see the differences. They'd likely need to use the diff tool to compare received and approved for any more.

loop-evgeny:
Raised a TeamCity issue for this, by the way: https://youtrack.jetbrains.com/issue/TW-69539

PR #760, opened 2021-02-02
trim large strings

### LINKED ISSUE #736: ShouldMatchApproved performance issues with large approvals
We're having performance issues when using ShouldMatchApproved in various places when the exception message is shown.

- Visual studio intellisense for the exception message.
- Build server when the output of test runs shows the exception message.

This is due to the `expectedValue` being shown in the exception, when the approval is very large.

Is there a way to remove the `expectedValue` from the exception message? And possibly the diff as we don't need it either?

Repo - https://github.com/distantcam/ShouldlyIssue736

SimonCropp:
fixes #736 #749


SimonCropp:
ping @distantcam

JosephWoodward:
Nice, looks good to me 👍

jnm2 (approved):

jnm2:
Thanks @SimonCropp!

PR #765, opened 2021-02-11
Updated shouldBe.md example for ShouldNotBe

Lisa3x3x3:
The "ShouldNotBe" example contained the "ShouldBe" example from above. Updated it to reflect how "ShouldNotBe" would look and what the exception would look like.


jnm2 (approved):
Thanks, this is very much appreciated!

PR #772, opened 2021-02-28
Improve user quality of life when debugging assertions in VS by not embedding the PDB

### LINKED ISSUE #769: Embedded PDB causes VS to jump to Shouldly internals instead of user test code when debugging assertions
This is caused by VS not following the "Just My Code" setting and treating any DLL that contains embedded PDB as the user's code.

We had the same problem at NUnit and decided to go the route of the new recommended snupkg symbol package format.

https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/nuget#symbol-packages
https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg
https://devblogs.microsoft.com/nuget/improved-package-debugging-experience-with-the-nuget-org-symbol-server/

jnm2:
Fixes #769

@JosephWoodward When publishing the .nupkg to a feed, just make sure that the .snupkg is in the same directory as the .nupkg file and the right thing will happen by default.

https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg
https://devblogs.microsoft.com/nuget/improved-package-debugging-experience-with-the-nuget-org-symbol-server/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ output
 test-results
 *.gem
 *crunch*.xml
-*.nupkg
+*.*nupkg
 *DotSettings*
 *DS_Store
 *ncrunch*
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,7 @@ build_script:
 
 test: off
 artifacts:
-- path: nugets\*.nupkg
+- path: nugets\*.*nupkg
 
 on_failure:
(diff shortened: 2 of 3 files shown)


jnm2:
@JosephWoodward This might matter when it's time to push to nuget.org. Are you currently the only one with the ability to do that?

slang25 (approved):
This makes sense, I think you are right for the most part people will not want to step into Shouldly. snupkg is a good fit here 👍

SimonCropp:
closing this and can re-open if it is still occurring and someone re-raises

PR #778, opened 2021-03-17
Fix docs build

mcon:
Build looks to be broken as a result of some docs changes - have added the correct files so the solution cleanly builds.


jnm2:
Planning to leave this PR open for a few minutes as a test of the PR validation check until I see it's working.

jnm2:
Ran into a problem (https://github.com/SimonCropp/MarkdownSnippets/issues/370) and I'm coming back to this now that I have an idea of what a workaround would look like.

jnm2:
Never mind, apparently there is no way to trigger a check for a PR without pushing to the PR branch, and I found a PR of my own that I can play with.

jnm2:
@mcon Thanks for getting this done!

PR #780, opened 2021-03-18
Bump timeout to make Windows CI happier.

mcon:
Windows CI seems to be crazy slow and hits this 2 second timeout - let's just bump it to 10 seconds.

Build https://ci.appveyor.com/project/shouldly/shouldly/builds/38274996 for https://github.com/shouldly/shouldly/pull/778 shows this should be the remaining barrier to getting the build green.


jnm2:
@mcon Thank you for the help! I hadn't seen these, but I recommend a firm minimum of 10 seconds based on experience with CI. :+1:

mcon:
> @mcon Thank you for the help! I hadn't seen these, but I recommend a firm minimum of 10 seconds based on experience with CI. 👍

I only saw this once I fixed the docs build which is probably why you hadn't seen them before 👍

Issue #755, opened 2021-01-21
ShouldBeEquivalentTo ignores fields


loop-evgeny:
v4.0.3. ShouldBeEquivalentTo does not appear to compare fields - it passes for two objects with different field values, like this:

```
    [TestFixture]
    public class SomeTests
    {
        private class MyObject
        {
            public string Field;
            public string Property { get; set; }
        }

        [Test]
        public void CompareObjects()
        {
            //new MyObject {Property = "actual"}.ShouldBeEquivalentTo(new MyObject {Property = "expected"}); // Shouldly fails - good
            new MyObject {Field = "actual"}.ShouldBeEquivalentTo(new MyObject {Field = "expected"}); // Shouldly - passes!
            new MyObject {Field = "actual"}.Should().BeEquivalentTo(new MyObject {Field = "expected"}); // FluentAssertions - fails as expected
        }
    }
```

I see no documentation on ShouldBeEquivalentTo on docs.shouldly.io but sounds like it's supposed to behave like the FluentAssertions equivalent?

JosephWoodward:
Hi @loop-evgeny,

Thanks for raising the issue you're seeing. This feature is still in its infancy and is rather primitive at the so feedback like this is super useful.

jnm2:
I'm calling this a bug because I think it was missed in design to consider how we wanted this to play out, and I'm tentatively milestoning for 4.1 to minimize the amount of time in which tests might be written that could break when we roll this out.

mcon:
Should be able to close this one now https://github.com/shouldly/shouldly/pull/790 is in :+1:

PR #790, opened 2021-03-23
Fields now supported by `ShouldBeEquivalentTo`

mcon:
As raised in https://github.com/shouldly/shouldly/issues/755 this PR adds support for comparison of fields by `ShouldBeEquivalentTo`.

I've run the modified tests with and without the production code change, and verified `ShouldFailWhenTitleDoesNotMatch` erroneously does not fail in that case.

### DIFF
--- a/src/Shouldly.Tests/ShouldBeEquivalentTo/ObjectScenario.cs
+++ b/src/Shouldly.Tests/ShouldBeEquivalentTo/ObjectScenario.cs
@@ -86,18 +86,18 @@ Additional Info:
         }
 
         [Fact]
-        public void ShouldFailWhenTitleDoesNotMatch()
+        public void ShouldFailWhenFieldDoesNotMatch()
         {
             var subject = new FakeObject
             {
                 Id = 5,
-                Title = "Mr",
+                TitleField = "Mr",
                 Name = "Bob",
             };
             var expected = new FakeObject
             {
                 Id = 5,
-                Title = "Sir",
+                TitleField = "Sir",
                 Name = "Bob",
             };
             Verify.ShouldFail(() =>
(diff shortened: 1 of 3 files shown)


src/Shouldly.Tests/ShouldBeEquivalentTo/ObjectScenario.cs
```diff
@@ -81,6 +81,51 @@ but was
     but was
 ""Bob""
 
+Additional Info:
+    Some additional context");
+        }
+
+        [Fact]
+        public void ShouldFailWhenTitleDoesNotMatch()
```
jnm2, line 89: Could you name this `ShouldFailWhenFieldDoesNotMatch` to describe the feature being tested?

jnm2, line 89: Could you name this `ShouldFailWhenFieldDoesNotMatch` to describe the feature being tested?

src/Shouldly.Tests/ShouldBeEquivalentTo/ObjectScenario.cs
```diff
@@ -81,6 +81,51 @@ but was
...
+            var subject = new FakeObject
+            {
+                Id = 5,
+                Title = "Mr",
+                Name = "Bob",
+            };
+            var expected = new FakeObject
+            {
+                Id = 5,
+                Title = "Sir",
```
jnm2, line 100: Could either name this `Field`/`TitleField` or initialize only this value so that it is easy to see by looking at the test code where the field is?

jnm2 (changes requested):
Looks nice, thanks!

jnm2 (approved):

jnm2:
Thanks for your help!

Issue #767, opened 2021-02-16
ShouldBeEquivalentTo does not handle nested dictionaries


loop-evgeny:
This basic nested dictionary case fails with `ShouldBeEquivalentTo`

```
        [Test]
        public void CompareNestedDictionaries()
        {
            Dictionary<string, Dictionary<string, string>> CreateDict()
            {
                var innerDict = new Dictionary<string, string> {{"inner", "value"}};
                return new Dictionary<string, Dictionary<string, string>> {{"outer", innerDict}};
            }

            CreateDict().ShouldBeEquivalentTo(CreateDict());
        }
```

I would expect this to pass, but it throws:

```
Shouldly.ShouldAssertException : Comparing object equivalence, at path:
CreateDict() [System.Collections.Generic.Dictionary`2[[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Collections.Generic.Dictionary`2[[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]]
    Element [0] [System.Collections.Generic.KeyValuePair`2[[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Collections.Generic.Dictionary`2[[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]]

    Expected value to be
["outer" => [["inner" => "value"]]]
    but was
["outer" => [["inner" => "value"]]]
```

terryaney:
Similarily, there is a problem with `ShouldContainKeyAndValue` when value is an array.

```
[Fact]
public async Task Shouldly_ContainKeyAndValue()
{
	var d = new Dictionary<string, string[]>
	{
[...]

PR #792, opened 2021-03-23
Add support for comparing dictionaries

mcon:
This should cover the primary issue highlighted as part of https://github.com/shouldly/shouldly/issues/767 - the failure output is a little verbose: could we consider eliding some of the type qualifications when the output relating to type specification reaches a certain size?

Maybe should also think about the error message when an expected key is missing, just saying the key was "null" at the moment isn't ideal.

### DIFF
--- a/src/Shouldly.Tests/ShouldBeEquivalentTo/DictionaryScenario.cs
+++ b/src/Shouldly.Tests/ShouldBeEquivalentTo/DictionaryScenario.cs
@@ -13,7 +13,7 @@ namespace Shouldly.Tests.ShouldBeEquivalentTo
         }
 
         [Fact]
-        public void ShouldFailWhenComparingNonIdenticalNestedDictionaries()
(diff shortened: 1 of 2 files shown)


jnm2:
I'm at a point where I'm comfortable with this kind of change yet. See https://github.com/shouldly/shouldly/issues/794#issuecomment-808941364. We would have to be ready to be consistent and take on every combination of dictionary keys and values, enumerable values, options, nullables, tuples, custom option and result types, soon task types, etc.

Right now Shouldly is consistent. It always delegates to object equality except for some special helpers at the very top level. I don't think we should start supporting things in a jagged fashion where `Dictionary<A, Dictionary<B, C>>` behaves this way but `ValueTuple<A, Dictionary<B, C>>` does not.

I'll move this to discussion in the issue, but I wanted to explain why we aren't ready to take a PR like this at this stage of the design thought.

I really appreciate your work on this!

PR #796, opened 2021-03-29
Add squiggles and light bulb fixes to maintain style with the least amount of time taken

### LINKED ISSUE #770: Cut down on PR noise and provide easy, fast feedback to contributors by adding *a few* StyleCop Analyzers rules
@shouldly/core Would you mind if I turn on a few rules provided by StyleCop? Having contributed myself to repositories that have these rules in place, it's really nice having inline prompts and <kbd>Ctrl</kbd>+<kbd>.</kbd> light bulb fixes so that I don't spend time thinking about the codebase style to guess it up front, or spend time talking about it in my PRs.

These are some of my pet peeves that I don't want to bug PR contributors about after the PR is already up:

```diff
-if(condition)
+if (condition)
```

```diff
     Foo();
-
 }
```

However, I strongly disagree with over half the StyleCop Analyzers rules, such as always requiring braces, so I would not be enabling rules that don't seem fairly universal. I'd provide a pull request with one commit per rule I enable so that we can drop any that worry you.

jnm2:
Closes #770

I said a few and then turned on half the rules. 😄

I'm happy to drop any commit or rule. The goal with each rule is to keep the codebase from having weird formatting issues that hurt readability or maintainability, while minimizing the total amount of time any future contributor ends up spending on such things between writing code and the PR stage. If a rule does not do that, or gets in the way of something we want to do even slightly, it should be deleted.


jnm2:
Oops, I need to update test string literals.

jnm2:
Forgot to merge the master branch to fix warnings from PRs that have been merged in the meantime but which don't cause Git conflicts. I'll fix this by pushing directly.

Issue #766, opened 2021-02-15
ShouldMatch regexPattern parameter should be annotated with [RegexPatternAttribute]


jhv2017kam:
This will enable Resharper to syntax highlight the supplied regex string

jnm2:
This makes sense to me. We need to be sure that we don't ship any new public types or any new nuspec dependencies. I can think of at least one easy way to do that (declare an internal source attribute).

PR #797, opened 2021-03-29
Add RegexPattern attributes to regex parameters

mcon:
Addresses https://github.com/shouldly/shouldly/issues/766: Shouldly already has the `Jetbrains.Annotations` attribute source as internal attributes - no further types are being exposed, and there's no need to depend on `Jetbrains.Annotations` through nuget.

I'm able to see that usages in `Shouldly.Test` are appropriately highlighted once the attribute is added.


jnm2 (approved):
Ah, thank you!

PR #799, opened 2021-03-29
Update ShouldBeginEndWithMessageGenerator.cs

Shayankamalzadeh:
Use ? operator instead of if-else statement.

### DIFF
--- a/src/Shouldly/MessageGenerators/ShouldBeginEndWithMessageGenerator.cs
+++ b/src/Shouldly/MessageGenerators/ShouldBeginEndWithMessageGenerator.cs
@@ -16,24 +16,8 @@ namespace Shouldly.MessageGenerators
             var codePart = context.CodePart;
             var expected = context.Expected.ToStringAwesomely();
             var actualValue = context.Actual.ToStringAwesomely();
-            string actual;
-            if (codePart == actualValue)
-            {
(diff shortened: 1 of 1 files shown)


src/Shouldly/MessageGenerators/ShouldBeginEndWithMessageGenerator.cs
```diff
@@ -16,28 +16,12 @@ public override string GenerateErrorMessage(IShouldlyAssertionContext context)
...
-                }
-            }
-            else
-            {
-                actual = $@" was
-{actualValue}";
-            }
-
+            var actual = codePart == actualValue
+                ? context.IsNegatedAssertion ? " did" : " did not" : $@" was {actualValue}";
```
jnm2, line 20: This removed the intentional newline that is part of the message, and I imagine this is the reason there are failing tests.

jnm2, line 20: This reads confusingly to me. When a line starts with `?`, could you please make its matching `:` also start a new line?

jnm2 (changes requested):

jnm2:
Closing due to lack of attention. Thank you for wanting to improve Shouldly, and we'd welcome another look whenever you're back.

Issue #818, opened 2021-06-29
Should.ThrowAsync<Exception> does not fail for a succeeding task


hwanders:
The assertion `Should.ThrowAsync<TException>` with the generic type `Exception` will **not** fail for a successful task.
Instead it returns  a `ShouldAssertException` which I think should never be returned but only thrown.

This means that some tests including assertions about certain processes to fail (with an unknown exception) will appear green although the processes succeed.

Changing the asserted exception type to any other one works: the succeeding tasks gets detected and the assertion fails.

For reference: xunit's `Assert.ThrowsAnyAsync<Exception>(...)` properly fails in this case.

Demo code for Shouldly 4.0.3 (with xunit 2.4.1 on Microsoft.NET.Test.Sdk 16.10.0 on net5.0 with C# 9.0):

```csharp
[Fact]
public async Task ShouldThrowAsyncException_SucceedingTask_ShouldFailAssertion()
{
    // for reference, this assertion passes
    await Should.NotThrowAsync(() => Task.CompletedTask);

    // assert the opposite
    var ex = await Should.ThrowAsync<Exception>(() => Task.CompletedTask);

    // assertion should have failed!
    throw new Exception(
        $"{nameof(Should.ThrowAsync)} has not complained about the task's success; " +
        $"instead it returned a {ex.GetType().Name}"
    );
}
```

adam-knights:
New to this project, but fancied having a dig in the code. It is down to [this line](https://github.com/shouldly/shouldly/blob/037a2b9c118aa266a84988d9adf0301c4ca1ad5a/src/Shouldly/ShouldStaticClasses/ShouldThrowTaskAsync.cs#L55). When your expectedException type is `Exception` the `ShouldAssertException` is a child of `Exception` and matches, when its any other type of Exception it hits the second part of the switch and throws it and behaves as expected.

I've posted a PR with a fix that also adds a test that fails before my fix and passes afterward.

hwanders:
I think you've analyzed this correctly and the test is correct.
The `e.ShouldBeOfType<ShouldAssertException>()` in the `catch (ShouldAssertException e)` block seems redundant though.

ggrondecki:
I just ran into this issue today and I see the PR for it is still open. Any ETA on when this will be released?

PR #822, opened 2021-07-24
Fix Should.ThrowAsync<Exception> does not fail for a succeeding task

adam-knights:
Fixes https://github.com/shouldly/shouldly/issues/818.

Although it's not best practice for applications to throw generic Exception's, nothing stops this and we can support it here with no pain by reordering our switch to go from specific downward

The test could be written a couple ways, but the key for it to be red before the change is the final `throw new XunitException("ShouldThrowAsync did not throw");`, which i understand is Xunit's equivalent of `Assert.Fail(message)`.

### DIFF
--- a/src/Shouldly.Tests/ShouldThrowAsync/FuncOfTaskScenarioAsync.cs
+++ b/src/Shouldly.Tests/ShouldThrowAsync/FuncOfTaskScenarioAsync.cs
@@ -2,6 +2,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
+using Xunit.Sdk;
 
 namespace Shouldly.Tests.ShouldThrowAsync
 {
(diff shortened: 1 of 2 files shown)


src/Shouldly.Tests/ShouldThrowAsync/FuncOfTaskScenarioAsync.cs
```diff
@@ -157,5 +158,27 @@ but threw
...
+
+        [Fact] // Issue 818
+        public async Task ShouldThrowAssertException_ExceptionTypeIsException()
+        {
+            try
+            {
+                Func<Task> doSomething = () => Task.CompletedTask;
+                await Should.ThrowAsync<Exception>(async () => await doSomething());
+            }
+            catch (ShouldAssertException e)
```
hwanders, line 170: I think it would be better to use `catch (Exception e)` here to catch unexpected errors and get an appropriate error message from the `ShouldBeOfType` assertion below.

adam-knights: Agreed and updated, it now matches the tests above, I think I had it leftover as ShouldAssertException from my initial testing.

hwanders (approved):

hwanders (approved):

adam-knights:
@jnm2 or @JosephWoodward any chance of a review please?

adam-knights:
@slang25 any chance of help getting this merged? Seems someone else commented on the original issue that they hit this too.

PR #827, opened 2021-10-03
update DiffEngine and EmptyFiles

### LINKED ISSUE #826: Shouldly should not add tons of dependencies
There's a ton of odd dependencies introduced by the upgrade from 3x to 4x
I get that DiffEngine and EmptyFiles probably makes sense, but things like

Registry, MEF, ConfigurationManager, SqlClient and all the other seem pretty weird. Was there any thought put into crosscompiling so there are less dependencies?

grafik
grafik

SimonCropp:
fixes #826


slang25 (approved):
Looks good 🙂

PR #839, opened 2022-04-28
Add upgrade doc example of how `ShouldNotBeNull` change could cause compile errors 

IanKemp:
Sadly this actually happened in a codebase I maintain, and was very confusing for a while!


SimonCropp:
thanks

PR #840, opened 2022-06-29
Optimistic cancellation support for CompleteIn

rick-palmsens:
This pull requests adds the following overloads for `CompleteIn`:

```csharp
public static void CompleteIn(Func<CancellationToken, Task> actual, TimeSpan timeout, string? customMessage = null);
public static T CompleteIn<T>(Func<CancellationToken, Task<T>> actual, TimeSpan timeout, string? customMessage = null);
```

This allows the user to properly cancel their task when there is a timeout. This is personally required to cancel a `Stream.ReadAsync` when a connected device is taking too long to respond during our integration tests.


SimonCropp:
@rick-palmsens it was easier to redo the Pr than handle the conflicts https://github.com/shouldly/shouldly/pull/862 (i added you as a co-author so u r still credited)

PR #859, opened 2022-09-09
Add support for ReadOnlyDictionary

### LINKED ISSUE #834: Add support for IReadOnlyDictionary
`ShouldContainKey` works with `IDictionary`, but not with `IReadOnlyDictionary`.

spencerr:
Resolves #834

### DIFF
--- a/src/Shouldly.Tests/Dictionaries/ShouldContainKey.cs
+++ b/src/Shouldly.Tests/Dictionaries/ShouldContainKey.cs
@@ -9,10 +9,10 @@ public class ShouldContainKey
     private void ClassScenarioShouldFail(IEnumerable<KeyValuePair<MyThing, MyThing>> classDictionary)
     {
         Verify.ShouldFail(() =>
-classDictionary.ShouldContainKey(new MyThing(), "Some additional context"),
+                classDictionary.ShouldContainKey(new MyThing(), "Some additional context"),
 
-errorWithSource:
-@"classDictionary
+            errorWithSource:
+            @"classDictionary
     should contain key
 Shouldly.Tests.TestHelpers.MyThing (000000)
     but does not
@@ -20,8 +20,8 @@ Shouldly.Tests.TestHelpers.MyThing (000000)
 Additional Info:
     Some additional context",
 
-errorWithoutSource:
-@"[[Shouldly.Tests.TestHelpers.MyThing (000000) => Shouldly.Tests.TestHelpers.MyThing (000000)]]
+            errorWithoutSource:
+            @"[[Shouldly.Tests.TestHelpers.MyThing (000000) => Shouldly.Tests.TestHelpers.MyThing (000000)]]
     should contain key
 Shouldly.Tests.TestHelpers.MyThing (000000)
     but does not
@@ -35,10 +35,10 @@ Additional Info:
     public void GuidScenarioShouldFail(IEnumerable<KeyValuePair<Guid, Guid>> guidDictionary)
     {
         Verify.ShouldFail(() =>
-guidDictionary.ShouldContainKey(_missingGuid, "Some additional context"),
+                guidDictionary.ShouldContainKey(_missingGuid, "Some additional context"),
 
-errorWithSource:
-@"guidDictionary
+            errorWithSource:
+            @"guidDictionary
     should contain key
 5250646b-4c46-4b0e-86d8-e1421f2a0ea2
     but does not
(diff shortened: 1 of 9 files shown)


SimonCropp:
@spencerr can you take a look at the CI error

src/Shouldly.Tests/Dictionaries/ShouldContainKey.cs
```diff
@@ -27,23 +30,24 @@ public void ClassScenarioShouldFail()
...
 
-    [Fact]
-    public void GuidScenarioShouldFail()
+    [Theory]
+    [MemberData(nameof(GuidDictionaries))]
+    public void GuidScenarioShouldFail(IEnumerable<KeyValuePair<Guid, Guid>> guidDictionary)
     {
         Verify.ShouldFail(() =>
-                _guidDictionary.ShouldContainKey(_missingGuid, "Some additional context"),
+guidDictionary.ShouldContainKey(_missingGuid, "Some additional context"),
```
SimonCropp, line 38: can u fix the indetationn

src/Shouldly/ShouldlyExtensionMethods/ShouldBeDictionaryTestExtensions.cs
```diff
@@ -6,31 +6,60 @@ namespace Shouldly;
 [ShouldlyMethods]
 public static partial class ShouldBeDictionaryTestExtensions
 {
-    public static void ShouldContainKey<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key, string? customMessage = null)
+    public static void ShouldContainKey<TKey, TValue>(this IEnumerable<KeyValuePair<TKey, TValue>> dictionary, TKey key, string? customMessage = null)
```
SimonCropp, line 9: this is unfortunately a breaking runtime change. both overloads need to exist for this to be deployed in a minor

jnm2: @SimonCropp I can't see the runtime break. Can you describe it? It might be palatable as long as it doesn't affect mainstream usage of `ShouldContainKey`.

Specifically, `IDictionary<TKey,TValue>` inherits `IEnumerable<KeyValuePair<TKey, TValue>>`. Anything that was passed previously in a method call will behave the same as before.

src/Shouldly/MessageGenerators/DictionaryShouldContainKeyAndValueMessageGenerator.cs
```diff
@@ -6,11 +6,11 @@ namespace Shouldly.MessageGenerators;
 
 internal class DictionaryShouldContainKeyAndValueMessageGenerator : ShouldlyMessageGenerator
 {
-    private static readonly Regex Validator = new("ShouldContainKeyAndValue");
+    private static readonly Regex Validator = new Regex("ShouldContainKeyAndValue");
 
     public override bool CanProcess(IShouldlyAssertionContext context)
     {
-        return Validator.IsMatch(context.ShouldMethod);
+        return Validator.IsMatch(context.ShouldMethod) && context.Actual is IDictionary;
```
jnm2, line 13: We should support classes that implement `IReadOnlyDictionary<,>` and do *not* implement the non-generic `IDictionary`.

jnm2 (changes requested):

src/Shouldly/ShouldlyExtensionMethods/ShouldBeDictionaryTestExtensions.cs
SimonCropp: at runtime consumers will bind to `ShouldContainKey(IDictionary)`. so if u have a 3rd party dll that is consuming `ShouldContainKey(IDictionary)`, and that dll is not recompiled, there will be a missing method exception

jnm2: Are you sure the runtime doesn't make things work dynamically at runtime? The reason I ask is that the very first thing in the .NET breaking changes rules says that this is specifically not a breaking change: https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md

jnm2: Okay, upon reading the whole first 'Allowed' entry under https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md, this would not be allowed. Huh, I wonder what bullets I've dodged 😆

martincostello:
I was about to open an issue asking for support for `IReadOnlyDictionary`, then saw this PR.

If you're not going to come back to this @spencerr, any objections to me opening a new PR that picks this up to move it forwards?

jnm2:
@martincostello It's safe to assume that this PR has stalled. We'd be grateful for your help, thanks for offering!

PR #862, opened 2022-09-14
Optimistic cancellation support for CompleteIn

SimonCropp:
Co-Authored-By: rick-palmsens <57390463+rick-palmsens@users.noreply.github.com>

### DIFF
--- a/src/Shouldly/ShouldStaticClasses/ShouldCompleteInExtensions.cs
+++ b/src/Shouldly/ShouldStaticClasses/ShouldCompleteInExtensions.cs
@@ -35,6 +35,22 @@ public static partial class Should
         return task.Result;
     }
 
+    /*** CompleteIn(Func<Task>) ***/
+    public static void CompleteIn(Func<CancellationToken, Task> actual, TimeSpan timeout, string? customMessage = null)
+    {
+        var cancellationTokenSource = new CancellationTokenSource();
+        CompleteIn(actual(cancellationTokenSource.Token), timeout, customMessage, "Task", cancellationTokenSource);
+    }
+
+    /*** CompleteIn(Func<Task<T>>) ***/
+    public static T CompleteIn<T>(Func<CancellationToken, Task<T>> actual, TimeSpan timeout, string? customMessage = null)
+    {
+        var cancellationTokenSource = new CancellationTokenSource();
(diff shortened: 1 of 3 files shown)


src/Shouldly/ShouldStaticClasses/ShouldCompleteInExtensions.cs
```diff
@@ -64,6 +80,25 @@ private static void CompleteIn(Task actual, TimeSpan timeout, string? customMess
...
+                        actual.Wait();
+                    }
+                    catch (AggregateException aggregateException)
+                    {
+                        if (aggregateException.InnerExceptions.Count != 1 ||
+                            aggregateException.InnerException is not OperationCanceledException canceledException ||
+                            canceledException.CancellationToken != cancellationTokenSource.Token)
+                        {
+                            throw;
+                        }
```
jnm2, line 98: This could become a catch filter. `catch (AggregateException ex) when (ex...`

src/Shouldly/ShouldStaticClasses/ShouldCompleteInExtensions.cs
```diff
@@ -64,6 +80,25 @@ private static void CompleteIn(Task actual, TimeSpan timeout, string? customMess
...
+                    cancellationTokenSource.Cancel();
+                    try
+                    {
+                        actual.Wait();
+                    }
+                    catch (AggregateException aggregateException)
+                    {
+                        if (aggregateException.InnerExceptions.Count != 1 ||
+                            aggregateException.InnerException is not OperationCanceledException canceledException ||
+                            canceledException.CancellationToken != cancellationTokenSource.Token)
```
jnm2, line 95: I'm worried about this. Plenty of times, what is thrown is a TaskCanceledException with no matching CancellationToken. I just observed this with Entity Framework last week.

jnm2: I'm also not sure what I think about the blocking wait in the first place. It could legitimately deadlock in a test with a message pump like NUnit installs if you put [Apartment(STA)] on an async method.

If it's okay to not wait for the task when not using cooperative cancellation, why is it not okay to not wait for the task either way?

SimonCropp:
closing this. CompleteIn should be made async

PR #863, opened 2022-10-07
Install .NET SDK in GitHub Actions

martincostello:
- Add action to install the .NET SDK if the one on the agent [doesn't match the one in `global.json`](https://github.com/shouldly/shouldly/actions/runs/3063640825/jobs/4945942905).
- Update GitHub Actions to their latest versions.


slang25 (approved):
Thanks!

Issue #860, opened 2022-09-10
http://shouldly.io/ seems 💀 


gsigler:
None of the links to http://shouldly.io/ seem to be working. Looks like something went wrong.

MorganVergara:
No one seems to be monitoring the Gitter chat page either. This was reported there back on August 12th, and again by me yesterday (September 22d).

jnm2:
@JakeGinnivan @josephwoodward @SimonCropp The official docs have been down for two months or more. Who is capable of configuring the domain and hosting?

jnm2:
According to a WHOIS search, the domain `shouldly.io` does not expire until Aug 2023. This means there must be an issue with the hosting.

The hosting is https://www.gitbook.com if I understand correctly. (See the mention in #669 for example.) I'm not sure who has access to configure GitBook.

jnm2:
Or, maybe the name servers aren't set up correctly in the domain configuration. (https://docs.gitbook.com/advanced-guides/custom-domain/configure-dns)

jnm2:
I grabbed the domain **shouldly.org** so that we have that as an option if we don't have access to **shouldly.io** for one reason or another. Maybe I could give it a week and see if we get a response from @JakeGinnivan and @josephwoodward, and if not, I could attempt to publish the docs to `shouldly.org` via GitBook.

If we do revive **shouldly.io**, I'll set up **shouldly.org** to redirect to it. Then let's have a conversation about placing *all* these credentials, domains and hosting and instructions for publishing, in the hands of the most available Shouldly maintainers.

jnm2:
Current situation:

- The shouldly.io domain registration has expired and has been bought by a domain squatter who wants to sell it for a high price.
- GitBook, the hosting service, is now locking us out because adding me as a member of the GitBook Shouldly org put us over the limit for a free ("OSS") account.
- The administrator with access to these things has extremely limited time.

I'm investigating setting up my own GitBook account and connecting it to `shouldly.org` which I bought.

jnm2:
https://docs.shouldly.org/ is now live! I'll work on changing links.

PR #864, opened 2022-10-09
Switch from shouldly.io to shouldly.org

jnm2:
Fixes https://github.com/shouldly/shouldly/issues/860


slang25 (approved):
Thanks for sorting out the new domain 🥳

PR #870, opened 2022-11-13
System.Half support

sungam3r:
https://devblogs.microsoft.com/dotnet/introducing-the-half-type/

@SimonCropp I found that I can't write simple assertion test for comparing `Half` types in GraphQL.NET.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 Shouldly is an assertion framework which focuses on giving great error messages when the assertion fails while being simple and terse.
 
 This is the old *Assert* way: 
-```csharp
+```cs
 Assert.That(contestant.Points, Is.EqualTo(1337));
 ```
 For your troubles, you get this message, when it fails:
@@ -17,7 +17,7 @@ For your troubles, you get this message, when it fails:
     Expected 1337 but was 0
 
 How it **Should** be:
-```csharp
+```cs
 contestant.Points.ShouldBe(1337);
 ```
 Which is just syntax, so far, but check out the message when it fails:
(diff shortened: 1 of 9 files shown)


SimonCropp (approved):

README.md
```diff
@@ -9,23 +9,23 @@
 Shouldly is an assertion framework which focuses on giving great error messages when the assertion fails while being simple and terse.
 
 This is the old *Assert* way: 
-```cs
+```csharp
```
Shane32, line 12: There are no reason for this change.  `cs` doesn't work in the GraphQL.NET project for the documentation, but this is a limitation of the GraphQL.NET documentation builder; `cs` works fine for GitHub markdown files.  I always use `cs` when possible.

src/Shouldly.Tests/ShouldBe/HalfScenarios.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    [Fact]
+    public void ShouldPass_SameInstance()
+    {
+        Half h = (Half)12345;
+        h.ShouldBe(h);
+    }
+
+    [Fact]
+    public void ShouldPass_DirrerentInstances()
```
Shane32, line 15:
```suggestion
    public void ShouldPass_DifferentInstances()
```

README.md
sungam3r: https://github.com/madskristensen/MarkdownEditor is the reason

изображение

изображение

Maybe there are others.

sungam3r: The same for https://marketplace.visualstudio.com/items?itemName=MadsKristensen.MarkdownEditor2

These two are the most popular extensions for markdown files
изображение

Shane32: Those are both the same; just an older and newer version.   https://github.com/madskristensen/MarkdownEditor2022/issues/80

sungam3r: Thanks for posting issue.

Issue #795, opened 2021-03-26
Installing Shouldly silently disables optimization and changes other properties


loop-evgeny:
I've run into a similar issue before (#723) but this wasted hours of my time again: performance tests suddenly started failing for no explicable reason, when the code being tested did not change. The problem turned out to be Shouldly's .props file disabling optimization:

```
<?xml version="1.0" encoding="utf-8"?>
<Project>
  <PropertyGroup>
    <DebugSymbols>true</DebugSymbols>
    <Optimize>false</Optimize>
    <DebugType>embedded</DebugType>
    <Deterministic>false</Deterministic>
    <DeterministicSourcePaths>false</DeterministicSourcePaths>
  </PropertyGroup>
</Project>
```

This can be verified at runtime by [checking for the DebuggableAttribute](https://stackoverflow.com/a/49832939/1536933) (and I'm adding a unit test for it now). With Shouldly referenced by my test project optimization is disabled. It also still changes DebugType and, apparently disables Deterministic builds and source paths.

Guys, wasting your users time in this way is really not cool. The project looks promising, but this sort of stuff has me very close to giving up on using it. You may have your reasons for settings those properties, but please do not set build properties without a *huge, impossible to ignore* warning to the user somewhere. If you cannot provide a huge warning then don't do it at all and let the user do it manually *if* that's truly needed for Shouldly to work. For my usage, it seems to work just fine without those. (Tested this by just referencing the DLL directly, without the NuGet package.)

jnm2:
I hear you. Huge, impossible-to-ignore warnings for the 99% case are a non-starter I think, but maybe it would be an acceptable tradeoff to put a smaller amount friction in the 99% path and not have Shouldly work out of the box, but fail with a test message that says "Set this csproj property: ...". I think even this would diminish the user experience significantly and provide a poor impression of Shouldly for first-time users, even leading to many more people having the experience of their time being wasted.

If there was an idea that wouldn't involve nags or friction in the vast majority of cases, we'd jump at it, but I'm honestly not sure there can be a solution that works well everywhere. Without such an option, I'm not comfortable making a decision on this without the total agreement of @JosephWoodward and @slang25.

jnm2:
> if that's truly needed for Shouldly to work.

A central draw of Shouldly is that it generates great error messages. It only generates great error messages if it can locate source files based on the stack trace when a test fails. We don't currently know of a way to obtain the original source file without setting these properties, but we can also try again.

[...]

PR #881, opened 2023-03-08
Deterministic build

iBrotNano:
The build action was broken by https://github.com/shouldly/shouldly/pull/870. I had to remove the net5.0 target framework to make it work. It is already out of support so it should be removed anyway.

This pull request targets #795.

The idea is to make the configuration in `build.props` conditional. The standard way to use Shouldly is by referencing it in a test project. It should be no big deal to alter the build configuration of those. But if Shouldly is used in a non-test project those settings can break their builds. By setting a property `<UseShouldlyInNonTestProject>true</UseShouldlyInNonTestProject>` in the CSPROJ the settings in `build.props` should be ignored.

This would address https://github.com/shouldly/shouldly/issues/795#issuecomment-809150457 without the need for multiple nugets.

It's an alternative to https://github.com/shouldly/shouldly/issues/795#issuecomment-879084212. Maybe more descriptive.

### DIFF
--- a/src/Shouldly/Shouldly.csproj
+++ b/src/Shouldly/Shouldly.csproj
@@ -1,5 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
+    <!-- Had to remove the .net5.0 target. https://github.com/shouldly/shouldly/pull/870 -->
     <TargetFramework>netstandard2.0</TargetFramework>
     <Optimize>false</Optimize>
     <DebugSymbols>true</DebugSymbols>
@@ -14,7 +15,7 @@
     <PackageProjectUrl>https://shouldly.org</PackageProjectUrl>
     <PackageLicenseExpression>BSD-2-Clause</PackageLicenseExpression>
     <RepositoryType>git</RepositoryType>
-    <RepositoryUrl>https://github.com/shouldly/shouldly.git</RepositoryUrl>
+    <RepositoryUrl></RepositoryUrl> <!-- Is set by Microsoft.SourceLink.GitHub -->
     <PackageReleaseNotes>https://github.com/shouldly/releases/tag/$(Version)</PackageReleaseNotes>
     <GeneratePackageOnBuild Condition="$(Configuration) == 'Release'">true</GeneratePackageOnBuild>
     <PackageOutputPath>..\..\nugets</PackageOutputPath>
--- a/src/Shouldly/Shouldly.csproj
+++ b/src/Shouldly/Shouldly.csproj
@@ -19,6 +19,7 @@
     <GeneratePackageOnBuild Condition="$(Configuration) == 'Release'">true</GeneratePackageOnBuild>
     <PackageOutputPath>..\..\nugets</PackageOutputPath>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <WarningsNotAsErrors>NU1505</WarningsNotAsErrors>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="EmptyFiles" Version="4.1.0" PrivateAssets="None" />
--- a/src/Shouldly/Shouldly.csproj
+++ b/src/Shouldly/Shouldly.csproj
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <!-- Had to remove the .net5.0 target. https://github.com/shouldly/shouldly/pull/870 -->
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net5</TargetFrameworks>
     <Optimize>false</Optimize>
     <DebugSymbols>true</DebugSymbols>
     <DebugType>embedded</DebugType>
--- a/src/Shouldly/Shouldly.csproj
+++ b/src/Shouldly/Shouldly.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net5</TargetFrameworks>
+    <TargetFramework>netstandard2.0</TargetFramework>
     <Optimize>false</Optimize>
     <DebugSymbols>true</DebugSymbols>
     <DebugType>embedded</DebugType>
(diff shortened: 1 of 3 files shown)


sungam3r:
> The build action was broken by https://github.com/shouldly/shouldly/pull/870

In what way?

> It is already out of support so it should be removed anyway.

Wait a bit. Why? Just turning off this TFM you throw out the whole added code under `#if NET5_0_OR_GREATER` condition.

src/Shouldly/Shouldly.csproj
```diff
@@ -14,7 +15,7 @@
     <PackageProjectUrl>https://shouldly.org</PackageProjectUrl>
     <PackageLicenseExpression>BSD-2-Clause</PackageLicenseExpression>
     <RepositoryType>git</RepositoryType>
-    <RepositoryUrl>https://github.com/shouldly/shouldly.git</RepositoryUrl>
+    <RepositoryUrl></RepositoryUrl> <!-- Is set by Microsoft.SourceLink.GitHub -->
```
sungam3r, line 18:
```suggestion
```

iBrotNano:
> > The build action was broken by #870
> 
> In what way?
> 
> > It is already out of support so it should be removed anyway.
> 
> Wait a bit. Why? Just turning off this TFM you throw out the whole added code under `#if NET5_0_OR_GREATER` condition.

This was the output of the build:

```shell
Warning: C:\Program Files\dotnet\sdk\7.0.200\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.EolTargetFrameworks.targets(28,5): warning NETSDK1138: The target framework 'net5' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy. [D:\a\shouldly\shouldly\src\Shouldly\Shouldly.csproj::TargetFramework=net5]
D:\a\shouldly\shouldly\src\Shouldly\Shouldly.csproj : error NU1505: Warning As Error: Duplicate 'PackageDownload' items found. Remove the duplicate items or use the Update functionality to ensure a consistent restore behavior. The duplicate 'PackageDownload' items are: Microsoft.NETCore.App.Ref [5.0.0], Microsoft.NETCore.App.Ref [5.0.0]. [TargetFramework=net5]

Build FAILED.
```

It seems that adding the TFM net5.0 adds Microsoft.NETCore.App.Ref implicitly. That causes `Warning As Error: Duplicate 'PackageDownload' items found.`. But you are right. It would throw out the code for the type `Half`. I guess this was the purpose to add net5 at all.

I could observe that builds i made locally only show the first warning. Not the `error NU1505: Warning As Error`. This is only shown in the GitHub Action. Something must be different in the build container. I found https://github.com/dotnet/sdk/issues/24747

It might depend on the NuGet version of the build sdk.

I think it would be best to add `<WarningsNotAsErrors>NU1505</WarningsNotAsErrors>` to show the warning but enforce that this specific warning it is not treated as an error and leave net5 as TFM for now. You could upgrade later to net6.0 and/or net7.0

src/Shouldly/Shouldly.csproj
iBrotNano: The only benefit is to enable SourceLink in forks automatically. Otherwise, it would be out of sync with the sources in the origin repository.

iBrotNano:
Now the build in appveyor failed after I habe added documentation about the flag in README.md. The failing test is `Shouldly.Tests.ShouldBe.ShouldBeUnique.ComparerScenario.ComparerNotEqualsShouldFail`. I can't see how this could be related to a change in README.md.

src/Shouldly/Shouldly.csproj
```diff
@@ -14,11 +14,12 @@
...
     <PackageReleaseNotes>https://github.com/shouldly/releases/tag/$(Version)</PackageReleaseNotes>
     <GeneratePackageOnBuild Condition="$(Configuration) == 'Release'">true</GeneratePackageOnBuild>
     <PackageOutputPath>..\..\nugets</PackageOutputPath>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <WarningsNotAsErrors>NU1505</WarningsNotAsErrors>
```
sungam3r, line 22:
```suggestion
    <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1505</WarningsNotAsErrors>
```

iBrotNano:
This seems not to work like expected. I am closing the PR until I found a solution.

iBrotNano:
Ok. I figured out that MSBuild evaluates build properties in a way that it is not as easy as i thought to make it configureable. I described the configuration options and fallbacks in READEME.md.

slang25:
I've had a look, and this looks good, however I'm at the conclusion the the `build.props` should never have been included. 

The `build.props` settings are too intrusive, and now more than ever given there is less .NET Framework about, and deterministic builds are becoming more popular. I think we just remove it, and get the docs updated to address the side-effects if the assertion message stack walking fails.

iBrotNano:
Hi, 

what i could read in https://github.com/shouldly/shouldly/issues/795 was that there are two scenarios. The first is what most of the users will do. Use Shoudly in a test project. https://github.com/shouldly/shouldly/issues/795#issuecomment-808939494 explains that the settings in the `build.props` are needed to generate messages. I could see, that it is not possible to navigate from the message in VS Test Explorer to the source code if `build.props` are missing.

The second scenario is that people want to use Shouldly in a project as a dependency. In this case `build.props` might overwrite their build configuration. I think this is an edge case and https://github.com/shouldly/shouldly/issues/795#issuecomment-808937752 points out the problem.

I stumbled over this because of exactly this. For me this solution would be good enough. I can use all features locally in Visual Studio. On the build server I circumvent the settings with a flag.

iBrotNano:
Hi, 

what i could read in https://github.com/shouldly/shouldly/issues/795 was that there are two scenarios. The first is what most of the users will do. Use Shoudly in a test project. https://github.com/shouldly/shouldly/issues/795#issuecomment-808939494 explains that the settings in the `build.props` are needed to generate messages. I could see, that it is not possible to navigate from the message in VS Test Explorer to the source code if `build.props` are missing.

The second scenario is that people want to use Shouldly in a project as a dependency. In this case `build.props` might overwrite their build configuration. I think this is an edge case and https://github.com/shouldly/shouldly/issues/795#issuecomment-808937752 points out the problem.

I stumbled over this because of exactly this. For me this would be good enouph.

slang25:
@iBrotNano could you give me your opinion on #883 

I'm leaning towards this because what's proposed in this PR is fine, if you already know what's gone wrong. I've had many reports of people spending hours of investigation to then find Shoudly being culprit, that's something I think we should be avoiding as much as possible.

slang25:
Resolved in version 4.2.1

PR #884, opened 2023-03-19
add new overload of ShouldBeEquivalentTo for include or exclude properties of equality

Ershad95:
add two overload of ShouldBeEquivalentTo  method for custom equality .
with these methods client can check equality with 2 or more types

### DIFF
--- a/src/Shouldly/ShouldlyExtensionMethods/ShouldBeEquivalentToExtension.cs
+++ b/src/Shouldly/ShouldlyExtensionMethods/ShouldBeEquivalentToExtension.cs
@@ -1,0 +1,68 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using Shouldly;
+
+namespace ShouldlyExtension
+{
+    public static class ShouldBeEquivalentToExtension
+    {
+        public static void ShouldBeEquivalentTo(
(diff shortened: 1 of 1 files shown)


michael-freidgeim-webjet:
@slang25 it will be good to improve this PR to make ShouldBeEquivalentTo as flexible as FA equivalent method, e.g. to address https://github.com/shouldly/shouldly/issues/1034#issuecomment-2595933285

cvetomir-todorov:
What is the status of this? Is there a reason it has been delayed for more than 1.5 years?

slang25:
This PR should have had a response, but please understand that this isn't a case of hit merge and make another release, I mean it _could be_, but there should be some consideration to how this feature is intended to work and evolve more generally.

Folks coming from FluentAssertions for example are more familiar with excluding properties for equivalency checks rather than a list of properties to include. We'd probably want an options parameter parameter too so that we can tweak over time how equivalency works (so we could have both include & exclude).

I'm currently doing some analysis of what people have requested, what we currently support, and what FluentAssertions support, and then we can come up with a broader plan for this functionality that hits the sweet spot.

Today plugging in the Shouldly equivalency functionality into the FluentAssertions test suite gives these results:
image

so there is a massive gap between how these things behave, and changing the current behaviour will break things. So I'll close this PR and create a new issue to brainstorm the direction of equivalency.

PR #897, opened 2023-04-21
Avoid Microsoft.CSharp dependency on net5.0

ViktorHofer:
Microsoft.CSharp.dll is part of the net5.0 shared framework and a package reference to Microsoft.CSharp is ignored. While this is true, consumers of Shouldly still download the dependency unnecessarily. Minimizing the dependency graph makes restore more efficient.


sharwell (approved):

slang25:
Nice one @ViktorHofer

ViktorHofer:
Sure 👍

PR #898, opened 2023-04-22
Improved deterministic build support

slang25:
Following on from #883, it turns out that as suspected there is a need to support building and running tests in separate steps while also using deterministic builds.

This PR adds:
* Refactoring deterministic build code to `DeterministicBuildHelpers`
* A method to check for paths which "look deterministic" if the build tooling fails for some reason. These paths are prefixed with `/_/`, `/_1/` etc.. This is used in the `ShouldMatchApproved` method, where an accurate resolved location is essential.
* New MSBuild targets which produce a file during build capturing the path maps and copying this to the output folder, then reading this value during VSTest and setting it as an environment variable.
* In the case where `dotnet test` is ran on it's own, a build will be performed but it will be "within" the VSTest target, so wont run before it. In this case we set the environment variable directly, as we know the tests will run later and in the same process tree.
* A `DisableShouldlyPathMaps` build property users can set to `true` to completely disable this behaviour, just in case.

### DIFF
--- a/src/DeterministicTests/DeterministicTests.csproj
+++ b/src/DeterministicTests/DeterministicTests.csproj
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\Shouldly\buildTransitive\Shouldly.targets" />
   <PropertyGroup>
     <TargetFramework>net7.0</TargetFramework>
     <GenerateAssemblyInfo>true</GenerateAssemblyInfo>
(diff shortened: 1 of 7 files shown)


slang25:
@sungam3r this should solve the errors you reported, I'll work on testing this on one of the repos you linked to

slang25:
I've recreated your issue on [AspNetCore.Diagnostics.HealthChecks](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/tree/master) @sungam3r, and have confirmed that this new package fixes the issue.

I'll leave this PR open for comments for a day, if there are no comments then I'll look to get a `4.2.1` release out 🙂

sungam3r:
Thanks.

PR #930, opened 2023-10-20
Add support for ImmutableArray for ShouldBeEquivalentTo

larsiver:
ShouldBeEquivalentTo now succeeds when comparing instances of `ImmutableArray<T>` with equivalent elements. The wrong assumption was that if a type is a value type, then it cannot also be of type IEnumerable, but that is actually the case for `ImmutableArray<T>`


slang25 (approved):
Very nice! Thanks for this contribution

PR #933, opened 2023-12-29
Add NRT annotation to ShouldBeAssignableTo

Shane32:
Adds NRT annotation so that the return value from `ShouldBeAssignableTo<T>` is known to be not-null if the value passed to it is not-null.


Shane32:
I've updated the api approval file so tests should pass now.

sungam3r (approved):

PR #937, opened 2024-03-06
Update to .NET 6, remove dependencies on EmptyFiles and DiffEngine

ddezinski:


src/Shouldly/Shouldly.csproj
```diff
@@ -16,17 +16,15 @@
     <PackageProjectUrl>https://docs.shouldly.org</PackageProjectUrl>
     <PackageLicenseExpression>BSD-2-Clause</PackageLicenseExpression>
     <RepositoryType>git</RepositoryType>
-    <RepositoryUrl>https://github.com/shouldly/shouldly.git</RepositoryUrl>
-    <PackageReleaseNotes>https://github.com/shouldly/shouldly/releases/tag/$(Version)</PackageReleaseNotes>
+    <RepositoryUrl>https://github.com/ddezinski/shouldly.git</RepositoryUrl>
```
sungam3r, line 19:
```suggestion
    <RepositoryUrl>https://github.com/shouldly/shouldly.git</RepositoryUrl>
```

src/Shouldly/Shouldly.csproj
```diff
@@ -16,17 +16,15 @@
...
+    <PackageReleaseNotes>https://github.com/ddezinski/shouldly/releases/tag/$(Version)</PackageReleaseNotes>
```
sungam3r, line 20:
```suggestion
    <PackageReleaseNotes>https://github.com/shouldly/shouldly/releases/tag/$(Version)</PackageReleaseNotes>
```

slang25 (changes requested):
Let's not remove core functionality without some kind of discussion.

ddezinski:
Sorry I did not mean to send a pull request to the shouldly branch, I'm working on my own fork.  I don't use the DiffEngine or EmptyFiles functionality and these extra packages just add additional work to keep them up to date.  It would be great if these packages were optional or if there were two NuGet packages one with for those who need them and one without for those of us who do not.

sungam3r:
I would also remove these dependencies.

SimonCropp:
here is the PR that added diffengine if u want to revert it. https://github.com/shouldly/shouldly/pull/644

slang25:
> here is the PR that added diffengine if u want to revert it. #644

I don't think it should be removed from the official Shouldly repo, however I am thinking it should be a separate package (that adopts the same TFM strategy as DiffEngine), and be enabled with a one-liner config.

This keeps the deps down, and doesn't could binary concerns like TFM support. I think this would keep most people happy.

SimonCropp:
what TFMs are you trying to maintain support for?

slang25:
`netstandard2.0` in this case, but in general they are different projects with different strategies on TFM support, which the dependency version management more difficult. We're not the only package in the .NET ecosystem with this problem of course.

SimonCropp:
> netstandard2.0 

but the consumers of shoudly are test apps. which have to target a runtime. for example nunit has dropped support for netstandard. and the v3 xunit will be dropping support for netstandard.

SimonCropp:
side note: no objections to splitting the shoudly package

src/Shouldly/Shouldly.csproj
```diff
@@ -16,17 +16,15 @@
...
     <GeneratePackageOnBuild Condition="$(Configuration) == 'Release'">true</GeneratePackageOnBuild>
     <PackageOutputPath>..\..\nugets</PackageOutputPath>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <ContinuousIntegrationBuild Condition=" '$(CI)' == 'true' ">true</ContinuousIntegrationBuild>
     <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1505</WarningsNotAsErrors>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="DiffEngine" Version="11.3.0" />
-    <PackageReference Include="EmptyFiles" Version="4.4.0" PrivateAssets="None" />
```
Shane32, line 29: 💯 👍 🥇

Shane32:
> > netstandard2.0
> 
> but the consumers of shoudly are test apps. which have to target a runtime. for example nunit has dropped support for netstandard. and the v3 xunit will be dropping support for netstandard.

I'm exceedingly thankful that this library targets .NET Standard 2.0.

src/Shouldly/Shouldly.csproj
sungam3r: So do I

SimonCropp:
replaced by https://github.com/shouldly/shouldly/pull/1081

PR #965, opened 2024-07-08
Switch from AppVeyor to GitHub Actions

slang25:
While AppVeyor works, I've been finding the build agents frustratingly slow, so am looking at switching to GHA.
We won't be able to actually test this until the workflow is updated in the default branch, so there may be follow up PRs.


SimonCropp:
no complaints here. but how do we burn the appveyor build feedback?

slang25:
I think we delete the `appveyor.yml` from the default branch, and that should be it as far as I understand. I've removed the check from the branch protection and will add this new CI checks instead.

sungam3r:
Also need to clear appveyor web hook.

PR #984, opened 2024-08-24
Support IReadOnlyDictionary<K, V>

### LINKED ISSUE #834: Add support for IReadOnlyDictionary

martincostello:
- Add assertion methods for `IReadOnlyDictionary<K, V>` using `net9.0` and later only so that `[OverloadResolutionPriority]` can be used to avoid ambiguity caused by types that implement both dictionary interfaces.
- Update to .NET 9 SDK and C# 13.

Resolves #834.

### DIFF
--- a/src/DocumentationExamples/DocumentationExamples.csproj
+++ b/src/DocumentationExamples/DocumentationExamples.csproj
@@ -1,7 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net9.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/DocumentationExamples/DocumentationExamples.csproj
+++ b/src/DocumentationExamples/DocumentationExamples.csproj
@@ -2,7 +2,6 @@
 
   <PropertyGroup>
     <TargetFramework>net9.0</TargetFramework>
-    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Shouldly.Tests/Shouldly.Tests.csproj
+++ b/src/Shouldly.Tests/Shouldly.Tests.csproj
@@ -1,9 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net8.0</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net8.0;net48</TargetFrameworks>
+    <TargetFrameworks>net9.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);net48</TargetFrameworks>
     <Optimize>false</Optimize>
-    <LangVersion>12.0</LangVersion>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
   <ItemGroup>
     <Compile Remove="**\*.approved.cs;**\*.received.cs" />
--- a/src/Shouldly.Tests/Shouldly.Tests.csproj
+++ b/src/Shouldly.Tests/Shouldly.Tests.csproj
@@ -3,7 +3,6 @@
     <TargetFrameworks>net9.0</TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);net48</TargetFrameworks>
     <Optimize>false</Optimize>
-    <LangVersion>latest</LangVersion>
   </PropertyGroup>
   <ItemGroup>
     <Compile Remove="**\*.approved.cs;**\*.received.cs" />
--- a/src/Shouldly/Shouldly.csproj
+++ b/src/Shouldly/Shouldly.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net5.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net5.0;net9.0</TargetFrameworks>
     <Optimize>false</Optimize>
     <DebugSymbols>true</DebugSymbols>
     <DebugType>embedded</DebugType>
@@ -23,6 +23,7 @@
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <ContinuousIntegrationBuild Condition=" '$(CI)' == 'true' ">true</ContinuousIntegrationBuild>
     <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1505</WarningsNotAsErrors>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="DiffEngine" Version="11.3.0" />
--- a/src/Shouldly/Shouldly.csproj
+++ b/src/Shouldly/Shouldly.csproj
@@ -23,7 +23,6 @@
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <ContinuousIntegrationBuild Condition=" '$(CI)' == 'true' ">true</ContinuousIntegrationBuild>
     <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1505</WarningsNotAsErrors>
-    <LangVersion>latest</LangVersion>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="DiffEngine" Version="11.3.0" />
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 ﻿{
   "sdk": {
-    "version": "8.0.303",
+    "version": "9.0.100-preview.7.24407.12",
     "allowPrerelease": true,
     "rollForward": "latestFeature"
   }
--- a/src/DeterministicTests/DeterministicTests.csproj
+++ b/src/DeterministicTests/DeterministicTests.csproj
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\Shouldly\buildTransitive\Shouldly.targets" />
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net9.0</TargetFramework>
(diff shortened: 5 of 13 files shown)


jnm2:
> Add assertion methods for Dictionary<K, V> to resolve overload ambiguity when using a concrete dictionary as it implements both interfaces.

This leaves other dictionary types such as ImmutableDictionary out in the cold, which will be noticed.

I may do a followup to add an internal declaration of .NET 9's [OverloadResolutionPriorityAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.overloadresolutionpriorityattribute?view=net-9.0) and use it to help with this, but that only comes into play if you set your langversion to 13 (currently, preview). The only docs I know of for this are the spec itself, <https://github.com/dotnet/csharplang/blob/main/proposals/overload-resolution-priority.md>. This is how it works:

image

We could also consider removing the IDictionary extension method or moving it to a different namespace. I think these days it should be rare to have a dictionary that does not implement IReadOnlyDictionary.

@slang25 @SimonCropp Do you have any thoughts?

jnm2:
Well, one important issue with moving or removing the extension method is, how do you then assert on something typed as the interface `IDictionary<,>`?

Is it worth hitting ImmutableDictionary at least, and a few more concrete types, or perhaps just wait for irritated feedback to roll in? :D

jnm2:
Anyone who does hit this issue could declare their own extension method for the dictionary class that we didn't think of, like ImmutableDictionary/ConcurrentDictionary/ReadOnlyDictionary/etc, and simply call the Shouldly extension method after casting to IReadOnlyDictionary. This may not be an obvious workaround, but we could add this to our docs.

Another option: We could add the new extension method under `#if NET9_0_OR_GREATER`. That way it won't cause disruption at all, because people are expected to use langversion 13 with .NET 9 (and to get a worse experience across the board if they don't). And then OverloadResolutionPriorityAttribute will work.

@martincostello Would you be able to benefit from this new extension method if it was only available when the test project targets .NET 9?

martincostello:
> This leaves other dictionary types such as ImmutableDictionary out in the cold, which will be noticed.

I hadn't considered things like `ImmutableDictionary`, `FrozenDictionary`, `ReadOnlyDictionary` etc. I could look to add overloads for those too if desired.

> Would you be able to benefit from this new extension method if it was only available when the test project targets .NET 9?

For me that wouldn't be a problem, as the app where I found the need for this extension is going to be running .NET 9 rc.1 next month (assuming no last-minute issues).

jnm2:
Okay, perhaps we start with using OverloadPriorityAttribute and only making this overload available in .NET 9 (need to add a net9.0 framework target for this), instead of adding overloads for concrete types, and then revisit as needed?

martincostello:
Sure, I can do that - how do you want to go about that given that the .NET 9 SDK is still pre-release until November?

jnm2:
As soon as the CI image has a .NET 9 preview version that contains the compiler update to recognize OverloadResolutionPriorityAttribute, we can update our global.json file with minimal hassle:

https://github.com/shouldly/shouldly/blob/50e00451de5c3ecba99aac232db329a6104f41fc/global.json#L1-L7

Then we need to change the langversion to preview, with a comment that we'll change it to 13 when .NET 9 ships:

https://github.com/shouldly/shouldly/blob/50e00451de5c3ecba99aac232db329a6104f41fc/src/Directory.Build.props#L4

Then add an additional `net9.0` to TargetFrameworks in Shouldly.csproj and Shouldly.Tests.csproj.

Then wrap this around the added overload, along with applying `[OverloadResolutionPriority(1)]`:

```cs
#if NET9_0_OR_GREATER
```

jnm2:
Ideally we'd choose the earliest preview version that is actually required in order to get the compiler recognition of OverloadResolutionPriorityAttribute. That would help with any contributors to the repo who might already have that older preview installed, as well as being the best chance of all three of these images having at least that preview version or later:

https://github.com/shouldly/shouldly/blob/50e00451de5c3ecba99aac232db329a6104f41fc/.github/workflows/CI.yml#L27

jnm2:
Okay. Through trial and error, I've confirmed that the compiler only starts recognizing OverloadResolutionPriorityAttribute in .NET 9 Preview 7, in spite of the attribute being [added to the core libraries in Preview 5](https://github.com/dotnet/core/blob/main/release-notes/9.0/preview/preview5/api-diff/Microsoft.NETCore.App/9.0-preview5_System.Runtime.CompilerServices.md).

<https://github.com/dotnet/core/blob/main/release-notes/9.0/preview/preview7/csharp.md#prioritize-better-overloads-with-overloadresolutionpriority-attribute>
Via <https://devblogs.microsoft.com/dotnet/dotnet-9-preview-7/>

This means that the global.json version should be `9.0.100-preview.7.24407.12`, and we'll just have to see if all the CI images have preview 7 yet. We could script them to install preview 7 also, or just be patient.

martincostello:
Why not just use actions/setup-dotnet to install the SDK? IMO that's best practice as it's deterministic and doesn't need to wait on GitHub to install anything. This approach even works for their daily builds from their CI.

My question was more geared towards merging - using a preview is trivial, but do you want main using prerelease software? Because otherwise this is unmergeable until November.

jnm2:
That sounds good too! Glad t9 know you're familiar already.

Yes, main using prerelease SDK is good with me. It's usually a safe bet, and there are options in the event of a rare surprise.

src/DocumentationExamples/DocumentationExamples.csproj
```diff
@@ -1,7 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net9.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
```
jnm2, line 5: I would prefer not to use a floating version since this means different things at different times on different machines. We could accidentally start using a new lang feature without explicitly intending to opt into depending on the latest tooling.

When using `preview` in the projects that need a specific language feature, there's no way around it, but that's a temporary state of things that we'll fix as soon as it's possible to replace `preview` with a concrete version and retain access to that feature. And important to me to add an XML comment above it explaining which language feature we're waiting on until we can move this to a concrete language version.

Also, I would prefer keeping all projects on the same langversion by only editing here, instead of adding LangVersion in any .csproj files:

https://github.com/shouldly/shouldly/blob/50e00451de5c3ecba99aac232db329a6104f41fc/src/Directory.Build.props#L4

jnm2: I did further testing and Preview 7 allows the use of langversion 13, so we don't need to set it to `preview` 🎉

Because of this, let's not touch langversion anywhere. The defaults should let us compile.

jnm2 (changes requested):
Great stuff! Thanks for the help. LGTM after reversing the langversion edits.

src/DocumentationExamples/DocumentationExamples.csproj
martincostello, line 5:
```suggestion
```

src/Shouldly.Tests/Shouldly.Tests.csproj
```diff
@@ -1,9 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net8.0</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net8.0;net48</TargetFrameworks>
+    <TargetFrameworks>net9.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);net48</TargetFrameworks>
     <Optimize>false</Optimize>
-    <LangVersion>12.0</LangVersion>
+    <LangVersion>latest</LangVersion>
```
martincostello, line 6:
```suggestion
```

src/Shouldly/Shouldly.csproj
```diff
@@ -23,6 +23,7 @@
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <ContinuousIntegrationBuild Condition=" '$(CI)' == 'true' ">true</ContinuousIntegrationBuild>
     <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1505</WarningsNotAsErrors>
+    <LangVersion>latest</LangVersion>
```
martincostello, line 26:
```suggestion
```

src/DocumentationExamples/DocumentationExamples.csproj
martincostello: AFAIK you'll only ever get a new C# version with a new major .NET version, so I don't think there's really much "accident" happening in that scenario. If you want repeatability global.json shouldn't roll-forward or allow previews so you always get the version that is in source control. The with Latest you don't have to remember to update it every year to get all the new shinies 😄

I didn't spot the D.B.props file as it wasn't in the root which is where I'm used to finding it, and there was a project-level setting so I just assumed that was the pattern here.

jnm2: I'm not quickly finding a link for it, but `latest` is discouraged by the C# language team. Rollforward of the SDK version makes the contribution experience so much smoother for both experienced folks and newer contributors. Since `latest` overrides the default language for the target framework when you have a later SDK, that's where the problem lies. Lastly, remembering to update langversion is usually not a hassle since the IDE prompts you to change it the moment you start to want it. And when no langversion is specified at all and the default is used based on TargetFramework, you don't have to update it at all unless you're after a preview feature.

I appreciate your thoughtfulness in approaching this!

jnm2 (approved):

jnm2:
@slang25, do you mind giving this a look and merging?

src/DocumentationExamples/DocumentationExamples.csproj
jnm2: I did find the link. The advice is buried underneath the table in the <https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/language#langversion> section:

> [!IMPORTANT]
> The `latest` value is generally not recommended. With it, the compiler enables the latest features, even if those features depend on updates not included in the configured target framework. Without this setting, your project uses the compiler version recommended for your target framework. You can update the target framework to access newer language features.

This did kick off a conversation. Bill plans to make this recommendation more visible soon.

jnm2:
@slang25 Next time I remember this PR exists, I'll probably merge it :D

slang25 (approved):
This looks good. I think we should separately update the `net5.0` tfm to `net8.0`, so that we have a policy of `netstandard2.0` + Current LTS + Current STS for the main package

PR #997, opened 2024-09-22
Drop .NET 5 and add .NET 8

martincostello:
- Drop `net5.0` TFM.
- Add `net8.0` TFM.

https://github.com/shouldly/shouldly/pull/984#pullrequestreview-2319929506

### DIFF
--- a/src/Shouldly.Tests/ShouldBe/HalfScenarios.cs
+++ b/src/Shouldly.Tests/ShouldBe/HalfScenarios.cs
@@ -1,4 +1,4 @@
-﻿#if NET5_0_OR_GREATER
+﻿#if NET8_0_OR_GREATER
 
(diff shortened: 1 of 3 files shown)


src/Shouldly.Tests/ShouldBe/HalfScenarios.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if NET5_0_OR_GREATER
+﻿#if NET8_0_OR_GREATER
```
jnm2, line 1: I like to keep these "or greater" checks as documentation for the exact .NET version in which the type appeared (5.0).

jnm2:
@martincostello Thanks so much!

@slang25 What do you think about another v4 release for this?

PR #1028, opened 2025-01-14
New ShouldHaveCount function

### LINKED ISSUE #270: New Assertion: ShouldCount
Following on from the [recent dicussion](https://github.com/shouldly/shouldly/issues/269) about adding more assertions, one I feel would be beneficial is `ShouldHaveCount` or `ShouldCount` (still undecided on which seems best).

Users can currently check to see if a collection is empty using `new int[0].ShouldBeEmpty(); but it would be great if they could check to see how many items are in the collection.

So I'm proposing something like this:

    new[] { 1,2,3 }.ShouldCount(3);

### LINKED ISSUE #365: ShouldHaveSingleItem()

twenzel:
Provides a `ShouldHaveCount' method for IEnumerals.

Fixes #270


documentation/documentation/enumerable/have.md
```diff
@@ -1,5 +1,30 @@
...
 
+## ShouldHaveCount
+
+<!-- snippet: EnumerableShouldHaveCountExamples.ShouldHaveCount.codeSample.approved.cs -->
+<a id='snippet-EnumerableShouldHaveCountExamples.ShouldHaveCount.codeSample.approved.cs'></a>
+```cs
+var maggie = new Person { Name = "Maggie" };
+var homer = new Person { Name = "Homer" };
+var simpsonsBabies = new List<Person> { homer, maggie };
+simpsonsBabies.ShouldHaveCount(3);
```
sungam3r, line 11:
```suggestion
simpsonsBabies.ShouldHaveCount(2);
```

sungam3r: Also Homer does not look like baby.

dougcunha (approved):

PR #1030, opened 2025-01-14
Provide simple As extension

twenzel:
Provides a simple `As<MyType>()' extension to don't break the fluent API.

So instead of
```csharp
((PersonList)myContainer.Result).ShouldContain("test");
````
you can write
```csharp
myContainer.Result.As<PersonList>().ShouldContain("test");
```

### DIFF
--- a/src/Shouldly.Tests/ExtensionsTests.cs
+++ b/src/Shouldly.Tests/ExtensionsTests.cs
@@ -1,0 +1,27 @@
+﻿namespace Shouldly.Tests;
+
+public class ExtensionsTests
+{
+    [Fact]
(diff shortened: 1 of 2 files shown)


slang25:
We already have `ShouldBeOfType` so can write:
```
myContainer.Result.ShouldBeOfType<PersonList>().ShouldContain("test");
```

I think this is probably more practical as the tempation with with a null returning method is to make further assertions with null propagation, so you won't get an exception when it's not of the type you expect, or if you omit the null propagation (like in your example) you'll get a null reference exception rather than something clearer.

twenzel:
@slang25 You're right. Missed that method. 👍

PR #1032, opened 2025-01-15
Removing previous maintainers in README to avoid confusion/spam

slang25:


recumbent:
The same details you removed from the top level README exist here: [documentation/README.md](https://github.com/shouldly/shouldly/blob/master/documentation/README.md) - don't want to tread on toes by pushing a PR 🙂

PR #1045, opened 2025-01-16
Add xunit v3 marker interfaces

### LINKED ISSUE #996: xUnit 3 Support
xUnit v3 has some conventions around assertion exceptions which we could adopt to make the experience better.

https://xunit.net/docs/getting-started/v3/whats-new

Specifically it would look for a couple of marker interfaces which indicate assertion failures (`IAssertionException`) and timeouts (`ITestTimeoutException`). These can be in any namespace the documentation says.

slang25:
Fixes #996 

I've tested locally with an ad-hoc test project. Not sure of a better way to test other than doing `dotnet run -- -reporter json`, which outputs:
```json
// ...
"Cause": "Assertion",
// ...
```
and 
```json
// ...
"Cause": "Timeout",
// ...
```
accordingly.

### DIFF
--- a/src/Shouldly/Internals/XunitV3Markers/IAssertionException.cs
+++ b/src/Shouldly/Internals/XunitV3Markers/IAssertionException.cs
@@ -1,0 +1,6 @@
+namespace Shouldly.Internals.XunitV3Markers;
+
+/// <summary>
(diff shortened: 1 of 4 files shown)


SimonCropp:
LGTM. added to the 4.3 milestone

PR #1049, opened 2025-01-20
Fix handling of zero tolerances

martincostello:
Fix assertions incorrectly failing when a tolerance of "zero" for the type is used.

I migrated Polly from FluentAssertions to Shouldly over the weekend and we had a test like this:

```csharp
value.Should().BeCloseTo(expected, TimeSpan.Zero); // value is a TimeSpan
```

Migrating to Shouldly, I changed this to:

```csharp
value.ShouldBe(expected, TimeSpan.Zero);
```

The test then failed despite the values being equal.

I guess the original author wanted to be explicit? This was easily resolved by just changing the assertion to `value.ShouldBe(expected)`.

This PR fixes that behaviour by changing `<` to `<=` for `TimeSpan`, as well as `decimal`, `double`, `DateTime` and `DateTimeOffset` that also have the same issue.

If non-zero tolerances being equal _shouldn't_ match, then an alternative fix could be to special-case "zero" for these cases in `Is`.


slang25:
Thanks @martincostello, the tolerance should be inclusive IMO, so this is bug, thanks for the fix

Issue #1074, opened 2025-02-07
TypeLoadException occures when using Verify in a project that has Shouldly installed


biohazard999:
Hello Everybody!

I've encountered a bug in combination when using Shouldy and Verify together in a project, I'm not sure if anyone encountered the same bug so I log it here.

System.TypeLoadException : Could not load type 'EmptyFiles.Extensions' from assembly 'EmptyFiles, Version=1.0.0.0, Culture=neutral, PublicKeyToken=c7a34512ecd69090'.

Step 1:
Create a simple xunit test project.
add Verify.XUnit (14.7.0)
and Shouldly (4.3.0)
packages.

add a simple test that uses Verify:

```cs
[UsesVerify]
public class ReproFacts
{
    [Fact]
    public async Task FailesWithTypeLoadException()
    {
        await Verifier.Verify("hello");
    }
}
```

I'll prepare little repro later. 

The PR #1071 already might solve the issue.

PR #1081, opened 2025-02-12
Move ShouldMatchApproved funtionality to Shouldly.Approvals

### LINKED ISSUE #939: DiffEngine compatibility issues
`DiffEngine` 14 dropped support for `Shouldly`'s current target frameworks `netstandard2.0` and `net5.0` and added a new optional `Encoding` argument for `DiffRunner.Launch`. The latest version of [`Verify`](https://github.com/VerifyTests/Verify) uses `DiffEngine` 15. This causes the following runtime error when referencing both `Shouldly` and `Verify` and calling `ShouldMatchApproved`:

```
System.MissingMethodException : Method not found: 'DiffEngine.LaunchResult DiffEngine.DiffRunner.Launch(System.String, System.String)'.
ShouldMatchApprovedTestExtensions.ShouldMatchApproved(String actual, Action`1 configureOptions, String customMessage)
```

I was initially going to submit a PR to bump DiffEngine but realized this resulted in more drastic changes. In order to support the new `DiffEngine` and resolve this issue, `Shouldly` would have to target `net472` and `net6.0` at a minimum (dropping `netstandard` altogether). Is this something you're willing to do?

SimonCropp:
fixes #939 and #1074


slang25 (commented):
These changes are looking good 😄 
I'll have a look at the failing tests, it's not picking up the source code in the error messages.

Let's move move `ShouldMatchApprovedException` into this the new project too.


On the library name, perhaps `Shouldly.Approvals`, although I'm happy with `Shouldly.Match`, just want to hear your thinking on this.

SimonCropp:
> perhaps Shouldly.Approvals

done

> move ShouldMatchApprovedException 

done

slang25:
I think moving the tests out into a `tests` dir at the top level should fix the build. There are things in the `Directory.Build.props` we want to isolate from the test projects (deterministic build for example)

src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
```diff
@@ -335,11 +335,6 @@ public class ShouldMatchApprovedException : Shouldly.ShouldAssertException
         public ShouldMatchApprovedException(string? message, string? receivedFile, string? approvedFile) { }
     }
     [Shouldly.ShouldlyMethods]
-    public static class ShouldMatchApprovedTestExtensions
```
sungam3r, line 338: Need to add new approved file?

src/Directory.Build.props
```diff
@@ -6,14 +6,35 @@
...
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <NoWarn>CS1591</NoWarn>
     <ImplicitUsings>enable</ImplicitUsings>
     <CheckEolTargetFramework>false</CheckEolTargetFramework>
     <DebugSymbols>true</DebugSymbols>
     <DebugType>embedded</DebugType>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(MSBuildProjectName)' == 'Shouldly.Approvals' Or '$(MSBuildProjectName)' == 'Shouldly'">
```
slang25, line 16: Nice

SimonCropp:
@sungam3r @slang25 so IMO this is ready to go. but it is a breaking change, so any opinions on when to ship?

one option is to ship a beta ASAP, so that unblocks people who this is a problem for

slang25 (approved):

sungam3r:
I am slightly watching the events taking place here. I'm fine with changes.

Shane32:
Is this going to remove .NET Standard support? This won’t work for the majority of my projects which test against older versions of the .NET Core framework.

slang25:
> Is this going to remove .NET Standard support? This won’t work for the majority of my projects which test against older versions of the .NET Core framework (as well as new versions).

For `ShouldMatchApproved` yes, but the rest of Shouldly will still target .NET Standard 2.0.

PR #1091, opened 2025-02-26
add AssertionScope

TalpalaruBogdan:
Add `AssertionScope` for Aggregated Error.

Sample Use:
```csharp
using (new AssertionScope())
{
    // Use the generic extension method for any assertion.
    actualInt.ShouldWithScope(x => x.ShouldBe(expectedInt));
    actualString.ShouldWithScope(x => x.ShouldBe(expectedString));
}
```

### DIFF
--- a/src/Shouldly.Tests/ShouldWithScope/ShouldThrowAggregateExceptionWithSingleAssertionScenario.cs
+++ b/src/Shouldly.Tests/ShouldWithScope/ShouldThrowAggregateExceptionWithSingleAssertionScenario.cs
@@ -1,0 +1,27 @@
+using Shouldly.ShouldlyExtensionMethods;
+
+namespace Shouldly.Tests.ShouldWithScope;
+
+public class ShouldThrowAggregateExceptionWithSingleAssertionScenario
+{
+    [Fact]
+    public void ShouldPass()
+    {
+        var actualString = "hello";
+        var expectedString = "world";
+
+        var action = new Action(() =>
+        {
+            using (new AssertionScope())
+            {
+                // Use the generic extension method for any assertion.
+                actualString.ShouldWithScope(x => x.ShouldBe(expectedString));
+            }
+        });
+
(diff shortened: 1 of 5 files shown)


slang25:
If we are to support assertion scopes, it should work without the additional `ShouldWithScope` methods, so the builtin assertion throw helpers would need to be aware of scopes.

slang25 (changes requested):
If we are to support assertion scopes, it should work without the additional ShouldWithScope methods, so the builtin assertion throw helpers would need to be aware of scopes.

DomasM:
@slang25 would this feature be added if implemented correctly?
```
using (new AssertionScope())
{
    actualInt.ShouldBe(expectedInt);
    actualString.ShouldBe(expectedString);
}
```

asos-gurpreetsingh:
> If we are to support assertion scopes, it should work without the additional ShouldWithScope methods, so the builtin assertion throw helpers would need to be aware of scopes.

This approach will be good to extend this to allow things like options.IgnoreProperty("Date")

src/Shouldly.Tests/ShouldWithScope/ShouldThrowAggregateExceptionWithSingleAssertionScenario.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        var actualString = "hello";
+        var expectedString = "world";
+
+        var action = new Action(() =>
+        {
+            using (new AssertionScope())
+            {
+                // Use the generic extension method for any assertion.
+                actualString.ShouldWithScope(x => x.ShouldBe(expectedString));
```
asos-gurpreetsingh, line 18: I think this should not use new wrapper method. The existing method should work with current assertion scope. same as fluent assertions does

DomasM:
there is open PR https://github.com/shouldly/shouldly/pull/1172 but it's stuck with no attention from maintainer

slang25:
As discussed in #1172, this feature is no longer planned. Thanks for the contribution anyway

Issue #1034, opened 2025-01-15
Onboarding guide for FluentAssertions


nicojs:
Hi 🙋‍♂️. I'm Nico Jansen, a manager at [Info Support](https://www.infosupport.com).

For years, we've sponsored [FluentAssertions](https://github.com/fluentassertions/fluentassertions), a comparable assertion library. My colleagues also contributed a lot of code to the project.  Unfortunately, they have decided to change the license to be a paid product for commercial usage; see https://github.com/fluentassertions/fluentassertions/pull/2943. As a response, we've decided to stop sponsoring FluentAssertsions immediately and sponsor you guys instead. We're delighted to help you create the best assertion library for .NET projects 🤝

By creating an onboarding guide, you would help us and, I'm sure, many others like us. What are the main differences between the projects, and how can you migrate from FluentAssertions to Shouldy? As this news travels around the internet, you can expect many more users to come in soon.

felipextrindade:
My project is migrating to Shouldly and we built a very simple "reference table" for the team. Hopefully this is useful in some way :)

|FluentAssertions|	Shouldly
|-- | --|
result.Should().Be(expected);	|result.ShouldBe(expected);
result.Should().NotBe(unexpected);	|result.ShouldNotBe(unexpected);
result.Should().BeNull();	|result.ShouldBeNull();
result.Should().NotBeNull();	|result.ShouldNotBeNull();
result.Should().BeOfType<SomeType>();	|result.ShouldBeOfType<SomeType>();
result.Should().NotBeOfType<SomeType>();	|result.ShouldNotBeOfType<SomeType>();
result.Should().BeGreaterThan(value);	|result.ShouldBeGreaterThan(value);
result.Should().BeLessThan(value);	|result.ShouldBeLessThan(value);
collection.Should().Contain(item);	|collection.ShouldContain(item);
collection.Should().NotContain(item);	|collection.ShouldNotContain(item);
collection.Should().BeEmpty();	|collection.ShouldBeEmpty();
collection.Should().NotBeEmpty();	|collection.ShouldNotBeEmpty();
collection.Should().HaveCount(count);	|collection.Count.ShouldBe(count);
action.Should().Throw<ExceptionType>();	|action.ShouldThrow<ExceptionType>();
action.Should().NotThrow<ExceptionType>();	|action.ShouldNotThrow<ExceptionType>();

felipextrindade:
[...]

PR #1094, opened 2025-02-27
Fix ShouldBeEquivalentTo using incorrect types on complex objects

nelsonroliveira:
# This PR fixes

Incorrect types used when comparing 2 instances of a class.

It was trying to use the type of the `actual` object instead of the type specified in the class, this caused issues with enumerable types that are frequently assigned in different but compatible ways.

This forces the use of the type declared for the property/field instead of relying on the type of the `actual` object.

Should fix the issues reported in https://github.com/shouldly/shouldly/issues/1034#issuecomment-2595206327

## Notes

The newly included test intentionally makes use of different kinds of objects while still respecting the types.

### DIFF
--- a/src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
+++ b/src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
@@ -33,6 +33,11 @@ namespace Shouldly
         public static TException Throw<TException>(System.Action actual, string? customMessage = null)
             where TException : System.Exception { }
     }
+    public class EquivalencyOptions
+    {
+        public EquivalencyOptions() { }
+        public bool CompareUsingRuntimeTypes { get; set; }
+    }
     public class ExpectedActualIgnoreOrderShouldlyMessage : Shouldly.ShouldlyMessage
     {
         public ExpectedActualIgnoreOrderShouldlyMessage(object? expected, object? actual, string? customMessage, [System.Runtime.CompilerServices.CallerMemberName] string shouldlyMethod = null) { }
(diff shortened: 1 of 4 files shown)


slang25 (changes requested):
This is a great start, in principle I'd like to get it merged.

Could you please add an optional options parameter to `ShouldBeEquivalentTo`, 
 using a new type `EquivalencyOptions`, which would have a `bool CompareUsingRuntimeTypes` property, this will default to the new behaviour in this PR (i.e. `false`), but allows users to switch back to the old behaviour.

We can then use this to add some other new features, such as ignore ordering, ignore properties, customizations etc... We'll never get close to the flexibility of FluentAssertions, but we can narrow the gap.

nelsonroliveira:
Done

src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
```diff
@@ -99,6 +104,7 @@ public interface IShouldlyAssertionContext
     public static class ObjectGraphTestExtensions
     {
         public static void ShouldBeEquivalentTo([System.Diagnostics.CodeAnalysis.NotNullIfNotNull("expected")] this object? actual, [System.Diagnostics.CodeAnalysis.NotNullIfNotNull("actual")] object? expected, string? customMessage = null) { }
+        public static void ShouldBeEquivalentTo([System.Diagnostics.CodeAnalysis.NotNullIfNotNull("expected")] this object? actual, [System.Diagnostics.CodeAnalysis.NotNullIfNotNull("actual")] object? expected, Shouldly.EquivalencyOptions options, string? customMessage = null) { }
```
asos-gurpreetsingh, line 107: Just a thought: The EquivalencyOptions or similar could be done the same way fluent assertions does AssertionScope but instead in shouldly we could have something like following. that way it wont be a breaking change

```
        using (var scope = new ShouldlyAssertionScope())
        {
            scope.IgnoreProperties("Date");

            actual.ShouldBeEqivalentTo(expected);
        }
```

slang25:
Heads up on scope, following the v5 equivalency roadmap in #1265: we've decided **not** to ship the `CompareUsingRuntimeTypes` flag in 5.0 after all — sorry for the earlier steer towards it. Options are purely additive, so if demand for runtime-type member selection shows up after the default flips we can add it in a 5.x (likely as an AOT-compatible closed-hierarchy variant rather than a bool).

What this PR remains (and it's still the foundation piece): the declared-type member-selection fix and the `EquivalencyOptions` seam itself. If you'd rather drop the options parameter entirely and keep this PR to just the declared-type fix, that works too — happy either way, and thanks again for sticking with this one.

slang25:
Superseded by #1266. The declared-type member selection and the `EquivalencyOptions` seam from this PR were cherry-picked into the v5 equivalency rewrite. Thanks @nelsonroliveira!

nelsonroliveira:
Glad to help, sorry for not answering sooner

Issue #1077, opened 2025-02-09
Can't assert equivalence of Dictionary<string, CustomObject>


rafa-ayadi-cloudm:
It does not seem to be an easy way to assert the equivalence of dictionaries with reference types.
Here is a simple example where the assertion fails:

`private class TestObj
{
    public int Id { get; set; }  
    public string Name { get; set; }    
}

  var dict1 = new Dictionary<int, TestObj>();
  dict1.Add(1, new TestObj() { Id = 1, Name= "Foo"});
  dict1.Add(2, new TestObj() { Id = 2, Name = "Bar" });
  var dict2 = new Dictionary<int, TestObj>();
  dict2.Add(1, new TestObj() { Id = 1, Name = "Foo"}); ;
  dict2.Add(2, new TestObj() { Id = 2, Name = "Bar" });

 dict1.ShouldBeEquivalentTo(dict2); //This fails

`

Any plans to support this? Or maybe a workaround?

wuzzeb:
Only workaround I have found is to convert to SortedDictionary.  ShouldBeEquivalentTo uses
the IEnumerable instance to compare but Dictionarys don't have a stable enumeration order.  Using SortedDictionary ensures that the keys are always enumerated in order so that the comparison will match.

The only annoying thing is that LINQ doesn't have a ToSortedDictionary method so have to use `new SortedDictionary<string, TestObject>(actual).ShouldBeEquivalentTo(expected)` in the test.

slang25:
Fixed by #1266: dictionaries now match by key and recurse into their values, so `Dictionary<string, CustomObject>` compares correctly. (Duplicate of the already-closed #1078.)

Issue #1075, opened 2025-02-07
Unable to Assert equivalence of 2 lists with complex objects with Ignoring order


rafa-ayadi-cloudm:
It doesn't look like there is a way to assert `ShouldBeEquivalentTo` for 2 lists of complex objects (Works with `ShouldBe(ignoreOrder: true)` for primitive types ).
Anyway we could get this implemented?

EliveltonRepolho:
Wanted to add some examples from my tests:

* This works as expected to [this unit test](https://github.com/shouldly/shouldly/blob/master/src/Shouldly.Tests/ShouldBe/EnumerableType/IgnoreOrderScenario.cs)
```C#
new List<int> { 1, 3, 2 }.ShouldBe([1, 2, 3], true, "Some additional context");
```

* This doesn't work, fail to compare considering ordering
```C#   
new
{
  Name = "Test",
  Items = new List<int> { 1, 3, 2 }
}.ShouldBeEquivalentTo(new
{
  Name = "Test",
  Items = new List<int> { 1, 2, 3 }
});
    
new
{
  Name = "Test",
  Items = new Dictionary<string, string>
  {
    { "Key_1", "Value_1" },
    { "Key_2", "Value_2" },
  } 
}.ShouldBeEquivalentTo(new
{
  Name = "Test",
  Items = new Dictionary<string, string>
  {
    { "Key_2", "Value_2" },
    { "Key_1", "Value_1" },
  } 
});
```

rafa-ayadi-cloudm:
> Best way I've found to do this is to sort the list first before passing to Shouldly. Then you know it's consistent. e.g. you don't care what order the objects are in.

My objects aren't really sortable, or at least there is no trivial way of sorting them.They could have the same values of properties, e.g same Ids.

slang25:
Fixed by #1266, which adds `EquivalencyOptions.IgnoreOrder` (with maximum bipartite matching) so lists of complex objects can be compared order-insensitively.

PR #1101, opened 2025-03-09
Feature: ISet & IDictionary Support For ShouldBeEquivalentTo()

Jarvis-Yu:
# Issue Addressed

- https://github.com/shouldly/shouldly/issues/1077
- https://github.com/shouldly/shouldly/issues/1075
- Also makes the behaviour closer to the equivalent in FluentAssertion

# New Behaviour

Supports `ISet` and `IDictionary` to be checked for equivalence correctly.

```c#
var setA = new HashSet<int>{1, 2, 3, 4};
var setB = new HashSet<int>{4, 3, 2, 1};
setA.ShouldBeEquivalentTo(setB);  // this passes

var dictionaryA = new Dictionary<int, int>{{1, 11}, {2, 222}};
var dictionaryB = new Dictionary<int, int>{{2, 222}, {1, 11}};
dictionaryA.ShouldBeEquivalentTo(dictionaryB);  // this passes
```

# Previous Behaviour

`ISet` and `IDictionary` was treated as `IEnumerable` which means order of elements affect how sets are checked.

```c#
var setA = new HashSet<int>{1, 2, 3, 4};
var setB = new HashSet<int>{4, 3, 2, 1};
setA.ShouldBeEquivalentTo(setB);  // this does not pass

var dictionaryA = new Dictionary<int, int>{{1, 11}, {2, 222}};
var dictionaryB = new Dictionary<int, int>{{2, 222}, {1, 11}};
dictionaryA.ShouldBeEquivalentTo(dictionaryB);  // this does not pass
```

### DIFF
--- a/src/Shouldly.Tests/ShouldBeEquivalentTo/DictionaryScenario.cs
+++ b/src/Shouldly.Tests/ShouldBeEquivalentTo/DictionaryScenario.cs
@@ -1,0 +1,251 @@
+using System.Text.RegularExpressions;
+
+namespace Shouldly.Tests.ShouldBeEquivalentTo;
+
+public class DictionaryScenario
+{
+    [Fact]
+    public void ShouldFailWhenDictionaryIsTooShort()
+    {
+        var subject = new Dictionary<int, int>
+        {
+            [1] = 2,
+            [2] = 3,
+            [3] = 5
+        };
+        var expected = new Dictionary<int, int>
+        {
+            [1] = 2,
+            [2] = 3,
+            [3] = 5,
+            [4] = 8
+        };
+        Verify.ShouldFail(() =>
+                subject.ShouldBeEquivalentTo(expected, "Some additional context"),
+
+            errorWithSource:
+            """
+            Comparing object equivalence, at path:
+            subject [System.Collections.Generic.Dictionary[[System.Int32],[System.Int32]]]
+                Keys
+
+                Expected value to be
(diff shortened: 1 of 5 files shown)


olivier-spinelli (commented):
Shouldn't this be based on `IReadOnlyDictionary` rather than `IDictionary`?
And `IReadOnlySet` rather than `ISet`?
For a maximal coverage, both Readonly and mutable interfaces should be tested (to allow pure immutable implementations to benefit from this good extension).

Jarvis-Yu:
> Shouldn't this be based on `IReadOnlyDictionary` rather than `IDictionary`? And `IReadOnlySet` rather than `ISet`? For a maximal coverage, both Readonly and mutable interfaces should be tested (to allow pure immutable implementations to benefit from this good extension).

Good suggestion. I additionally added support for `IReadOnlyDictionary` and `IImmutableSet`.

There is something funky going on stopping me from supporting `IReadOnlySet`, will check later on the details.

(Why `IImmutableDictionary` implements `IReadOnlyDictionary`, but `IImmutableSet` does not implement `IReadOnlySet`?)

By the way, `IDictionary` does not implement `IReadOnlyDictionary`, same for `ISet` and `IReadOnlySet`.

slang25:
Subsumed by #1266, which adds first-class set and dictionary comparison to ShouldBeEquivalentTo. The set and dictionary scenarios from this issue were folded into the new engine's test suite.

PR #1109, opened 2025-03-20
Support .NET Standard 2.0 for Shouldly.Approvals

Shane32:
Restores .NET Standard 2.0 compatibility for Shouldly.Approvals

See:
- #1081

### DIFF
--- a/src/Shouldly.Approvals/Shouldly.Approvals.csproj
+++ b/src/Shouldly.Approvals/Shouldly.Approvals.csproj
@@ -1,15 +1,16 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net472;net8.0;net9.0</TargetFrameworks>
+    <TargetFrameworks>net472;net8.0;net9.0;netstandard2.0</TargetFrameworks>
     <SignAssembly>true</SignAssembly>
   </PropertyGroup>
-  <ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' != 'netstandard2.0' ">
     <PackageReference Include="EmptyFiles" Version="8.8.0" PrivateAssets="None" />
     <PackageReference Include="DiffEngine" Version="15.10.1" />
     <None Include="..\..\assets\logo_128x128.png" Pack="true" PackagePath="assets" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-    <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
+    <PackageReference Include="EmptyFiles" Version="4.4.0" PrivateAssets="None" />
+    <PackageReference Include="DiffEngine" Version="11.3.0" />
     <PackageReference Include="TunnelVisionLabs.ReferenceAssemblyAnnotator" Version="1.0.0-alpha.160" PrivateAssets="all" />
     <PackageDownload Include="Microsoft.NETCore.App.Ref" Version="[8.0.0]" />
   </ItemGroup>


Shane32:
FYI, I am a maintainer and/or contributor for the following open-source projects, each of which utilize Shouldy's approval testing functionality on older frameworks:

- https://www.nuget.org/packages/QRCoder
- https://www.nuget.org/packages/Shane32.EasyPDF
- https://www.nuget.org/packages/GraphQL
- https://www.nuget.org/packages/GraphQL.AspNetCore3
- https://www.nuget.org/packages/GraphQL.Server.Transports.AspNetCore

I would prefer to continue to use the Shouldly approval library, however, all the above projects test against both old and current frameworks, and as such require targeting .NET Standard 2.0.

I also extensively use the Shouldly approval testing for internal projects.  However, most of these (but not all) target .NET 8.

SimonCropp:
@Shane32 but do the tests need to target  .NET Standard 2.0. ?

Shane32:
The test project itself must target a specific framework implementation and cannot target .NET Standard directly. The api approvals don't actually need to run under more than a single framework and .NET 8 is sufficient.  All of the above projects use ShouldMatchApproved for api approvals. In addition, 2 of the 5 listed above utilize ShouldMatchApproved within the unit tests which run across many frameworks.

SimonCropp:
> The test project itself must target a specific framework implementation and cannot target .NET Standard directly

so why does a test framework need to target  .NET Standard  ?

Shane32:
So it can be used with any framework.

SimonCropp:
@Shane32 what is the project with the best example of this requirement?

Shane32:
GraphQL.NET uses `ShouldMatchApproved` in several places.  Those within the main test project run on .NET Core 3.1, 6.0, 7.0 and 8.0, plus .NET Framework 4.8.  Those within the Harness test project run on .NET 6.0, 7.0 and 8.0.

https://github.com/search?q=repo%3Agraphql-dotnet%2Fgraphql-dotnet%20ShouldMatchApproved&type=code

Shane32.EasyPDF also uses `ShouldMatchApproved` for several tests.  The test project runs on .NET Core 2.1, 3.1, 6.0, 8.0 and .NET Framework 4.6.2, 4.7.2 and 4.8.

https://github.com/search?q=repo%3AShane32%2FEasyPDF%20ShouldMatchApproved&type=code

slang25:
It's still valid to target .NET Standard 2.0 today, so I don't think we should be putting extra constraints on our users. Microsoft have lots of code that still needs to target ns2.0 because of how MSBuild is hosted and that whole ecosystem, and @Shane32 examples above are completely valid too.

Perhaps you'd consider adding ns2.0 back @SimonCropp to these dependencies, it doesn't mean you support runtimes you don't want to support, but it doesn't force opinions downstream.

If not then I think we should merge this PR, but it seems like we're kicking the can down the road.

SimonCropp:
@slang25 

> Microsoft have lots of code that still needs to target ns2.0 because of how MSBuild is hosted and that whole ecosystem

yes, but AFAIK, no msbuilt task needs to ref 

it seems to me that @Shane32 actually wants support for  .NET Core 2.1, 3.1. both of which are out of support from MS and asking for netsandard2.0 is a way of achieving that. For me, as a FOSS maintainer,  providcing support for a framework that MS no longer supports, is waht i am trying to avoid

Shane32:
@slang25 Thank you.  I'd like to mention another option as well:

You could include approval support in the base Shoudly library, and refactor Shouldly.Approvals to be optional DiffEngine support, perhaps named Shouldly.DiffEngine.  This is a breaking change, as now a user may need to manually enable the diffengine (e.g. `Shouldly.ApprovalDefaults.Diff = Shouldly.DiffEngine.Instance;` or `.Diff()` or something) but the fact that the code is split into two libraries is already a breaking change.

With the above solution, Shouldly's entire assertion framework is still tied to .NET Standard 2.0 and lives in the main library, while DiffEngine support is now more like a plugin.  As a plugin it could either target .NET Standard in the same fashion as this PR, or it could only target the frameworks supported by the underlying libraries.

I would be happy to provide a PR demonstrating such a change if you wish.  (Personally I would prefer this approach.)

Shane32:
@slang25 close this now that there's very little code left in Shouldly.DiffEngine, or rebase?

PR #1111, opened 2025-03-22
Move approvals to main project with DiffEngine as a plugin

Shane32:
This PR moves all of the approval functionality back into the Shouldly project with the exception of the DiffEngine functionality.  Shouldly.DiffEngine now contains a few lines of code to tie `ShouldMatchApproved` to the DiffEngine package.

See discussion in #1109 

Pros:
- Users can use approval functionality without taking on the DiffEngine and EmptyFiles dependencies if they are not needed, allowing faster build times and simplifying configuration
- Users can write their own diff engine implementation easily and/or tie to a specific version of DiffEngine
- Shouldy.DiffEngine can be tied to the target frameworks of DiffEngine, if desired
- For users that do not use DiffEngine, there would be no changes required from prior versions (meaning, Shouldly.DiffEngine does not need to be installed)

Cons:
- Users are required to configure the diff engine as a default if they wish to use it, besides just installing the package
- Users may assume that the diff functionality was removed or not realize what additional setup is required

```cs
// in a static constructor that gets loaded before any tests
ShouldMatchConfiguration.ShouldMatchApprovedDefaults.ConfigureDiffEngine();

// or use .Diff() for each test to opt-in
```

Other options within this general design:
- Throw an exception when the diff engine is enabled but not configured with a message that explains how to properly configure it.  Perhaps also disable the diff engine by default.


src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
```diff
@@ -65,11 +65,28 @@ public class ExpectedShouldlyMessage : Shouldly.ShouldlyMessage
...
+    }
     [Shouldly.ShouldlyMethods]
     public static class GuidShouldBeTestExtensions
     {
         public static void ShouldBeEmpty(this System.Guid actual, string? customMessage = null) { }
     }
+    public interface IDiffEngine
+    {
+        void Launch(string receivedFile, string approvedFile);
+    }
```
Shane32, line 89: new

src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
```diff
@@ -330,6 +351,47 @@ public class ShouldContainWithCountShouldlyMessage : Shouldly.ShouldlyMessage
...
+    public static class ShouldMatchApprovedTestExtensions
+    {
+        public static void ShouldMatchApproved(this string actual, System.Action<Shouldly.ShouldMatchConfigurationBuilder>? configureOptions = null, string? customMessage = null) { }
+    }
+    public class ShouldMatchConfiguration
+    {
+        public ShouldMatchConfiguration() { }
+        public ShouldMatchConfiguration(Shouldly.ShouldMatchConfiguration initialConfig) { }
+        public string? ApprovalFileSubFolder { get; set; }
+        public Shouldly.IDiffEngine? DiffEngine { get; set; }
```
Shane32, line 368: new

src/Shouldly.Tests/ConventionTests/ApprovePublicApi.DiffEngineApi.approved.cs
```diff
@@ -0,0 +1,12 @@
...
+    public class DiffEngine : Shouldly.IDiffEngine
+    {
+        public static Shouldly.DiffEngine Instance { get; }
+        public void Launch(string receivedFile, string approvedFile) { }
+    }
+    public static class ShouldMatchConfigurationExtensions
+    {
+        public static Shouldly.ShouldMatchConfiguration ConfigureDiffEngine(this Shouldly.ShouldMatchConfiguration configuration) { }
+    }
+}
```
Shane32, line 12: new

src/Shouldly.Tests/ConventionTests/ApprovePublicApi.DiffEngineApi.approved.cs
```diff
@@ -0,0 +1,13 @@
...
+    {
+        public static Shouldly.DiffEngine Instance { get; }
+        public void Launch(string receivedFile, string approvedFile) { }
+    }
+    public static class ShouldMatchConfigurationBuilderExtensions
+    {
+        public static Shouldly.ShouldMatchConfigurationBuilder ConfigureDiffEngine(this Shouldly.ShouldMatchConfigurationBuilder builder) { }
+        public static Shouldly.ShouldMatchConfigurationBuilder Diff(this Shouldly.ShouldMatchConfigurationBuilder builder) { }
+    }
+}
```
Shane32, line 13: new

README.md
```diff
@@ -47,5 +47,15 @@ Alternatively if you're using .NET Core then you can install Shouldly via the co
 dotnet add package Shouldly
 ```
 
+To have `ShouldMatchApproval` display a diff of the expected and actual files, you will need to install the `Shouldly.DiffEngine` package and configure it.
+
+```bash
+Install-Package Shouldly.ShouldlyShouldMatchApproved
```
sungam3r, line 53:
```suggestion
Install-Package Shouldly.DiffEngine
```

src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
slang25: I really like this overall approach and think it's the way to go.

My only thought is calling it `IDiffEngine` feels accidentally coupled to the implication.
Can I suggest either:
* `IDiffViewer`
* `IDiffToolLauncher`
* ... something else, I'm welcome to suggestions

Shane32: I agree.  I also like `IDiffViewer` but whatever you think is fine.  Should I try that?  Let me know and I'll change the relevant property, interface, package name, folder, etc.

Shane32: Actually I would probably leave the package name alone, since it is tied to DiffEngine

src/Shouldly/ShouldMatchApprovedTestExtensions.cs
```diff
@@ -38,9 +42,9 @@ public static void ShouldMatchApproved(this string actual, Action<ShouldMatchCon
 
         if (!File.Exists(approvedFile))
         {
-            if (!config.PreventDiff)
+            if (!config.PreventDiff && config.DiffEngine != null)
```
slang25, line 45: I worry that for the 80% common case, folks will update without reading the release notes, and perhaps we need to be a bit more aggressive with telling folks here "Hey, we didn't launch a diff tool, see here for how to enable this" linking off to docs with the DiffEngine package instructions.

If this message popped up in a failing CI/CD step then I don't think it would be relevant, so a crude "is this an interactive environment" could be used here too, not a deal breaker though.

src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
slang25: Yeah, keep `Shouldly.DiffEngine` 👍 

Let's go with `IDiffViewer` and `DiffViewer` for the config property 🙂

src/Shouldly.DiffEngine/DiffEngine.cs
```diff
@@ -0,0 +1,15 @@
+using DiffEngine;
+
+namespace Shouldly;
+
+public class DiffEngine : IDiffViewer
```
Shane32, line 5: Should this be in the Shouldly.DiffEngine namespace and be called DiffViewer ?

slang25: I would say this is the `DiffEngineDiffViewer`, but tbh don't feel that strongly,

Shane32: Done

src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
Shane32: Done

src/Shouldly/ShouldMatchApprovedTestExtensions.cs
Shane32: > I worry that for the 80% common case, folks will update without reading the release notes, and perhaps we need to be a bit more aggressive with telling folks here "Hey, we didn't launch a diff tool, see here for how to enable this" linking off to docs with the DiffEngine package instructions.

Makes sense to me.

> If this message popped up in a failing CI/CD step then I don't think it would be relevant, so a crude "is this an interactive environment" could be used here too, not a deal breaker though.

So, you're thinking perhaps to throw if no DiffViewer is configured -- but only in interactive environments?  How do you suggest we do that?  This is ChatGPT's suggestion:

```cs
public static class CiDetector
{
    public static bool IsRunningOnCi()
    {
        // Check the common universal variable
        string ci = Environment.GetEnvironmentVariable("CI");
        if (!string.IsNullOrEmpty(ci) && 
            (ci.Equals("true", StringComparison.OrdinalIgnoreCase) || ci.Equals("1")))
        {
            return true;
        }

        // Check provider-specific environment variables
        if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("TEAMCITY_VERSION")) ||
            !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("TRAVIS")) ||
            !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("JENKINS_URL")) ||
            !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("GITLAB_CI")) ||
            !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("CIRCLECI")) ||
            !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("APPVEYOR")))
        {
            return true;
        }

        return false;
    }
}
```

slang25: I'm thinking not throwing a new exception, but extending the existing throw message with something to the effect of "There is no diff viewer configured, if you want to automatically view the relevant diff then ... (instructions)".

That snippet is the kind of thing I had in mind, it would include Azure Pipelines and GitHub Actions too.

Shane32: Good idea

Shane32: How about this?

```
Shouldly.ShouldMatchApprovedException : To approve the changes run this command:
copy /Y "C:\Users\shanek\source\repos\Shane32\shouldly\src\Shouldly.Tests\ConventionTests\ApprovePublicApi.ShouldlyApi.received.cs" "C:\Users\shanek\source\repos\Shane32\shouldly\src\Shouldly.Tests\ConventionTests\ApprovePublicApi.ShouldlyApi.approved.cs"

No diff viewer configured; if you want to automatically view the relevant diff,
please install Shouldly.DiffEngine and configure it via
ShouldMatchConfiguration.ShouldMatchApprovedDefaults.ConfigureDiffEngine();
----------------------------

publicApi
    should match approved with options: Ignoring line endings
"namespace Shouldly
{
    public class ActualFilteredWithPredicateShouldlyMessage : Shouldly.ShouldlyMessage
    {
        public ActualFilteredWithPredicateShouldlyMessage(System.Linq.Expressions.Expression filter, object? result, object? actual, string? customMessage, [System.Runtime.CompilerServices.CallerMemberName] string shouldlyMethod = null) { }
    }
    public class ActualShouldlyMessage : Shouldly.ShouldlyMessage
    {
```

Shane32: As programmed now, applying NoDiff() to the default configuration will suppress the 'no diff viewer' message.

src/Shouldly.Tests/ConventionTests/ApprovePublicApi.ShouldlyApi.approved.cs
```diff
@@ -330,6 +351,47 @@ public class ShouldContainWithCountShouldlyMessage : Shouldly.ShouldlyMessage
     {
         public ShouldContainWithCountShouldlyMessage(object? expected, object? actual, int matchCount, string? customMessage, [System.Runtime.CompilerServices.CallerMemberName] string shouldlyMethod = null) { }
     }
+    public class ShouldMatchApprovedException : Shouldly.ShouldAssertException
+    {
+        public ShouldMatchApprovedException(string? message, string? receivedFile, string? approvedFile, bool viewerNotPresent) { }
```
Shane32, line 356: `bool viewerNotPresent` added.  Can add prior constructor for non-breaking change if desired.

Shane32:
@slang25 Anything you want me to do on this PR?

slang25:
Hey @Shane32 , not at the moment, I've been busy with travel and am now also ill, but in a few days I'll get back to reviewing this.

Shane32:
No problem and sorry to hear that; get well soon!

slang25 (approved):

PR #1112, opened 2025-03-23
Add xml comments to public members

Shane32:


Shane32:
This PR adds XML comments to all public members of the Shouldly project.  After PR #1111 is either merged or rejected, I'll make a PR for the DiffEngine-related public members.  In a third PR, I'll add XML comments for the public classes and modify the csproj to fail the build when XML comments are missing from public API members in the future.

The comments are pretty generic but should be accurate; the project's member names generally describe the method quite well.  Please suggest changes as desired, and particularly for any members that would benefit from additional remarks within the XML comment.

src/Shouldly/ShouldlyExtensionMethods/ShouldBe/StringStartEndTestExtensions.cs
```diff
@@ -19,19 +25,28 @@ public static void ShouldEndWith([NotNull] this string? actual, string expected,
     }
 
     /*** ShouldNotStartWith ***/
```
slang25, line 27: This doesn't have to be part of this PR, but feel free to delete these `/*** coments ***/`

Shane32: Will do in a separate PR so as to not create merge conflicts and for better PR separation.

PR #1113, opened 2025-03-23
Remove byte order mark (BOM) from files

Shane32:
The editorconfig file currently specifies that BOM is not to be used when saving files (`charset = utf-8`).  However, a majority (not all) of cs files still have a BOM.  This PR removes the BOM from the remaining files in the repo.  This unifies the codebase and provides better compatibility as follows:
- With VS Code, as it occasionally duplicates the BOMs at the start of the file when using certain tools
- With VS, as it will always save in the format specified by the editorconfig


Shane32:
The api approval files was not changed.  I think it incorrectly contains BOM characters but I left it alone.

Besides cs files, BOMs were removed from:
- src/Shouldly.Tests/.editorconfig
- global.json
- src/mdsnippets.json

sungam3r (approved):

slang25 (approved):
Thanks for this

Shane32:
No problem :+1:

slang25:
Your 3 open PRs are looking good, do you have a merge order preference? If the order matters, I know logically they don't overlap.

Shane32:
No I don’t think so. If there’s a merge conflict, I’ll fix it.

PR #1132, opened 2025-04-16
Fixes ShouldThrowAsync exception formatting

GingerSnap-xx:
It was not formatted into a block quote like the rest of the documentation. 

Also moved the Include to be in the correct location.


GingerSnap-xx:
I'm not super familiiar with github workflows. Is there something I need to do on my fork side or something I need to do here in the mainline repo to get through these 3 pending checks?

PR #1151, opened 2025-08-20
Feat/should contain same elements as

vbandi:
Adds a new assertion to compare sequences for same elements while ignoring order, with optional comparer support.
Routes ignoring-order failures to the ignoring-order message generator for clearer diffs.
Adds unit tests including duplicates, nulls, lazy sequences, and numeric specials.
Updates the approved public API snapshot.


vbandi:
Sorry, wrong repo

PR #1152, opened 2025-08-27
Added ShouldNotThrow<TException>

eholman:
Added ShouldNotThrow<TException> to assert that the exception isn't thrown. 

I'm not quite sure if my contribution meets all requirements, it's my first one to any open source project! Any feedback is welcome


slang25:
Thanks @eholman — this fills a genuine gap and we want it for v5. I've rebased your commit onto master and pushed a follow-up commit (via maintainer edit access) with one deliberate semantic change plus the v5 mechanical updates.

**Semantic change: exceptions of other types now propagate instead of being swallowed.** The original implementation caught and ignored anything that wasn't `TException`, which means `Should.NotThrow<ValidationException>(action)` would stay green if the code started throwing `NullReferenceException` instead. `ShouldNotThrow<TException>` can honestly be read either way, so we resolved the ambiguity in the safe direction: the assertion only makes a claim about `TException`; any other exception propagates unchanged and still fails the test with its original stack trace. (FluentAssertions' `NotThrow<T>` swallows here — we're deliberately not matching that, since it can hide real bugs.)

The failure message now also names the disallowed type, including when a derived exception is caught:

```
`action`
    should not throw
System.SystemException
    but threw
System.InvalidOperationException
    with message
"the operation failed"
```

Also in the follow-up commit: v5-style `[CallerArgumentExpression]` signatures, propagation/derived-type tests, the regenerated public API approval file, and a documentation section. Async parity (`Should.NotThrowAsync<TException>`) is left as a follow-up.

Issue #1150, opened 2025-08-09
ShouldHaveSingleItem assertion message misleading


slang25:
When calling `.ShouldHaveSingleItem()` on a value that is nullable, it will say that there is a count of `0` when the value is `null`, it will then say that the value was `null`, but the description is confusing (at least to me 😓).

Example:
```csharp
string[]? nullableCollection = null;
nullableCollection.ShouldHaveSingleItem();
```
output:
```
nullableCollection
    should have single item but had
0
    items and was
null
```

GenjiruSUchiwa:
Hello, maybe something like can be less confusing ?

```csharp
string[]? nullableCollection = null;
nullableCollection.ShouldHaveSingleItem();
```

output:

```
nullableCollection should have a single item but it was null
```

Maybe it can be flagged as  `first timer` and i'll be happy to take it.

slang25:
Fixed by #1163, which is now merged — a null collection now reports "should have single item but was null" instead of "but had 0 items".

PR #1163, opened 2026-01-01
Fix ShouldHaveSingleItem message when collection is null (#1150)

luizgag:


slang25:
Thanks @luizgag — the fix itself still applies cleanly and the #1150 bug is still present on master, so we want this in.

The PR predated a v5 test-infrastructure change: `Verify.ShouldFail` no longer takes `errorWithSource`/`errorWithoutSource` — failure messages are now verified against `.approved.txt` files. That's why CI failed to compile. I've pushed an update to your branch (via maintainer edit access) converting `NullScenario` to the current style; your message-generator fix is untouched.

PR #1243, opened 2026-06-23
Fix ShouldContain throwing NullReferenceException on a null string

### LINKED ISSUE #821: ShouldContain throws exception on null input (string)
Consider code below:
```
//act
var result = GetResult();
//assert
result.ShouldContain("test string");
```
When `result` is null then the test will end up with exception: NullReferenceException.
In my opinion it should resulted with failing assertion, like:
```
Shouldly.ShouldAssertException : result
    should contain (case insensitive comparison)
"test string"
    but was actually
null
```

danielchalmers:
Fixes #821 - `ShouldContain`/`ShouldNotContain` throwing a raw `NullReferenceException` on a null string.

Use the null-conditional `actual?.Clip(...)` so a null actual flows through the normal message pipeline.


slang25:
Thanks @danielchalmers

PR #1256, opened 2026-07-03
Add FluentAssertions migration guide

danielchalmers:
Adds a Migrating from FluentAssertions guide which targets Shouldly v5 and addresses shouldly/shouldly#1034.

Every mapping and behavioural claim was checked against the v5 public API (`ApprovePublicApi.ShouldlyApi.approved.cs`) and executed against a local Shouldly v5 build alongside FluentAssertions. Dogfooded by migrating 21 real assertion patterns mined from MudBlazor's test suite (16k+ `.Should()` call sites); each migrated assertion matches FluentAssertions' pass/fail outcome on both passing and failing inputs.

### DIFF
--- a/documentation/SUMMARY.md
+++ b/documentation/SUMMARY.md
@@ -7,6 +7,7 @@
 ## Documentation
 
 * [Getting Started](documentation/getting-started.md)
+* [Migrating from FluentAssertions](documentation/migrating-from-fluentassertions.md)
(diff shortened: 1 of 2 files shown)


danielchalmers:
This was largely automated by Claude (Opus 4.8 + Fable 5) and needs further work, but I wanted to see if this is the kind of thing you are looking for before committing more time to it. Feel free to close if not!

PR #1285, opened 2026-07-18
Fix C# 12 ambiguity for collection ShouldBe calls

### LINKED ISSUE #1278: `ShouldBe(collection)` is ambiguous (CS0121) under `LangVersion < 13` in v5
## Summary

In v5, `someEnumerable.ShouldBe(expectedArray)` — the canonical collection-equality assertion — fails to compile with **`CS0121` (ambiguous call)** for any project whose effective `LangVersion` is **below 13**. The upgrade is clean for consumers on C# 13+, and silently breaks for those pinned to an older `LangVersion` (a common, stability-driven choice). Not mentioned in the [4 → 5 migration guide](https://github.com/shouldly/shouldly/blob/master/documentation/documentation/upgrade/4to5.md).

Found while dogfooding `5.0.0-preview.2` against real repositories.

## Symptom

```csharp
IEnumerable<Type> actual = things.Select(x => x.Type);
actual.ShouldBe(new[] { typeof(A), typeof(B) });
```

```
error CS0121: The call is ambiguous between the following methods or properties:
  'Shouldly.ShouldBeTestExtensions.ShouldBe<T>(T?, T?, string?, string?)' and
  'Shouldly.ShouldBeTestExtensions.ShouldBe<T>(IEnumerable<T>?, IEnumerable<T>?, bool, string?, string?)'
```

Triggers when `actual` is typed as `IEnumerable<T>` (e.g. a lazy LINQ result), so both the scalar and the enumerable `ShouldBe<T>` overloads are applicable.

## Root cause

v5's enumerable overload (`src/Shouldly/ShouldlyExtensionMethods/ShouldBe/GenericShouldBeTestExtensions.cs`) is decorated with `[OverloadResolutionPriority(1)]` to win over the scalar `ShouldBe<T>(T?, T?, …)`:

```csharp
[OverloadResolutionPriority(1)]
public static void ShouldBe<T>(this IEnumerable<T>? actual, IEnumerable<T>? expected, bool ignoreOrder = false, ...)
```

`[OverloadResolutionPriority]` is **only honored by the compiler when `LangVersion >= 13`**. Under an older `LangVersion` the attribute is ignored, the two overloads are equally applicable, and the compiler reports `CS0121`.

## Regression proof + LangVersion correlation

Minimal repro (file-based app):

```csharp
#:package Shouldly@5.0.0-preview.2
#:property LangVersion=10
IEnumerable<Type> actual = new[]{typeof(int)}.Select(t=>t);
actual.ShouldBe(new[]{typeof(int)});   // CS0121 at LangVersion 10; compiles fine at default (C# 14)
```

Across the repos we upgraded, it tracked `LangVersion` exactly:

| Repo | LangVersion | RE-3 fires? |
| --- | --- | --- |
| JasperFx/Wolverine | 12 | **yes — 12 call sites** |
| JasperFx/Lamar | 10 | yes — 1 site (compiled clean on 4.2.1) |
| JasperFx/Marten | 13 | no |
| JasperFx/Weasel | 13 | no |
| JasperFx/Alba | default (C# 14) | no |
| GitTools/GitVersion | latest | no |

The same code compiled cleanly on 4.x — this is a v5 regression, gated on `LangVersion < 13`.

## Workaround

Name the `bool` so only the enumerable overload is applicable (works at any LangVersion):

```csharp
actual.ShouldBe(expected, ignoreOrder: false);
```

…or raise `LangVersion` to 13+.

## Suggested fix

Prefer not to rely solely on `[OverloadResolutionPriority]` to disambiguate a core overload set, since it's silently a no-op below C# 13. Options: give the enumerable overload a signature that doesn't collide with the scalar one, or, at minimum, document the `LangVersion >= 13` requirement prominently in the migration guide.

AmirTahan80:
## Summary

Fixes collection `ShouldBe` calls that are ambiguous for consumers compiling with C# 12 or earlier.

Fixes #1278.

## Root cause

Shouldly uses `OverloadResolutionPriority` to prefer the enumerable comparison overload over the generic scalar overload. That priority is honored by C# 13 and newer compilers, but older compilers ignore the metadata. As a result, a call such as:

```csharp
IEnumerable<Type> actual = GetTypes();
actual.ShouldBe(new[] { typeof(int) });
```

fails with `CS0121` because the compiler cannot choose between the scalar and enumerable overloads.

## Changes

- Add an `IReadOnlyCollection<T>` expected-value overload for enumerable comparisons.
- Forward the new overload to the existing ordered enumerable comparison implementation.
- Preserve caller argument expressions and nullable annotations.
- Add a regression test covering an `IEnumerable<Type>` actual value compared with an array.
- Update the approved public API snapshot.

The more-specific expected type lets older compilers select collection comparison for common expected values such as arrays, lists, sets, and collection expressions without changing existing comparison behavior.

## Validation

- Verified the original reproduction fails in a standalone C# 12 consumer before the change and compiles after the change.
- Ran the full repository pipeline with `./build.ps1`.
- Release build and NuGet packaging passed.
- All 1,995 tests passed:
  - Shouldly.Tests: 1,833
  - EquivalencyComparisonTests: 73
  - DocumentationExamples: 89


slang25:
Wow that was fast @AmirTahan80 😅 I've already merged a fix to this, but thank you for getting involved

PR #1084, opened 2025-02-18
Feature/should be superset of

spaceorc:
Implement ShouldBeSupersetOf test extension, analogue to ShouldBeSubsetOf.
Example usage:

```csharp
new[] {1, 2, 3}.ShouldBeSupersetOf([2, 3]);
```

PR #1288, opened 2026-07-18
Add ShouldBeSupersetOf

### LINKED ISSUE #1085: Extension ShouldBeSupersetOf is absent
There is ShouldBeSubsetOf test extension, but pair ShouldBeSupersetOf is absent.
Please see a PR with the implementation (which is pretty much a fixed copy of ShouldBeSubsetOf implementation): https://github.com/shouldly/shouldly/pull/1084

danielchalmers:
Closes #1085 by adding `ShouldBeSupersetOf(expected)` + an `IEqualityComparer<T>` overload as the counterpart to `ShouldBeSubsetOf`. On failure the message lists the expected elements missing from `actual`, computed with the same comparer the assertion used.

This picks up #1084 by spaceorc, which was closed without being merged, and includes the docs page, a doc example and a FluentAssertions migration row (`Contain(IEnumerable)` → `ShouldBeSupersetOf`).

The missing elements reach the message generator on an internal context property rather than in place of `actual`, so a message that falls back to printing the actual value (no call-site expression) still shows the whole collection. #1290 uses the same property for `ShouldBeSubsetOf`, and the two merge cleanly in either order.


danielchalmers:
Note `ShouldBeSubsetOf` has the same defect as fd7aa68 in mirror form: the comparer overload asserts with the supplied comparer, but the generator recomputes the "outside subset" list with default equality https://github.com/shouldly/shouldly/blob/b3f3fee68f21c2dfbb897bba3199597cc6c24dd1/src/Shouldly/MessageGenerators/ShouldBeSubsetOfMessageGenerator.cs#L17

For example `new[] { "A", "Z" }.ShouldBeSubsetOf(["a"], StringComparer.OrdinalIgnoreCase)` fails only because of `"Z"` but reports both `"A"` and `"Z"` as outside the subset.

Split out as #1290 to keep this one scoped to `ShouldBeSupersetOf`.

danielchalmers:
`ShouldBeSupersetOf` does the same check as `ShouldBeSubsetOf` with the arguments swapped `expected.ShouldBeSubsetOf(actual)`, comparer overload included), and #1085 hasn't drawn much demand, so it isn't worth the extra API surface and review time.

PR #1310, opened 2026-08-03
Detect reference cycles through collections, sets, and dictionaries

### LINKED ISSUE #1307: `ShouldBeEquivalentTo` throws `StackOverflowException` on self-referential collections, dictionaries, and sets
## Summary

`ShouldBeEquivalentTo` guards against reference cycles only when the cycle closes through a **complex (member-wise) object**. When a cycle instead closes through a **collection**, **array**, **set**, or **dictionary** node, the comparison recurses forever and the process dies with an uncatchable `StackOverflowException` - which cannot be caught, so it crashes the whole test run/host.

## Steps to reproduce

Minimal console repro:

```csharp
using Shouldly;

var actual = new List<object>();
actual.Add(actual); // a list that contains itself

var expected = new List<object>();
expected.Add(expected); // structurally the same

actual.ShouldBeEquivalentTo(expected); // StackOverflowException
```

The same crash happens for other collection-shaped cycles:

```csharp
// Self-referential array
var a = new object[1];
a[0] = a;
var b = new object[1];
b[0] = b;
a.ShouldBeEquivalentTo(b); // StackOverflowException

// Self-referential HashSet
var sa = new HashSet<object> { 1 };
sa.Add(sa);
var sb = new HashSet<objec> { 1 };
sb.Add(sb);
sa.ShouldBeEquivalentTo(sb); // StackOverflowException

// Self-referential dictionary value
var da = new Dictionary<string, object> { ["value"] = 1 };
da["self"] = da;
var db = new Dictionary<string, object> { ["value"] = 1 };
db["self"] = db;
da.ShouldBeEquivalentTo(db); // StackOverflowException
```

## Expected behavior

The cycle should be detected and the two graphs treated as equivalent (they have identical shape), mirroring how Shouldly already handles a cycle through a plain object member. This case **works today**:

```csharp
class Node { public string Name = "x"; public Node? Next; }

var actual = new Node();
actual.Next = actual;
var expected = new Node();
expected.Next = expected;

actual.ShouldBeEquivalentTo(expected); // passes
```

So the object-graph cycle is handled, but the equivalent collection/dictionary/set cycle is not.

## Actual behavior

`StackOverflowException` - the runtime terminates the process immediately. Because it is uncatchable, a single such assertion aborts the entire test session (all other tests in the run are lost), rather than failing one test.

Bafyn:
Fixes #1307 

## Problem

`ShouldBeEquivalentTo` throws an uncatchable `StackOverflowException` when the compared graphs contain a reference cycle that closes through a **collection, array, set, or dictionary** node.

## Root cause

Cycle tracking (`_visitedPairs`) was only consulted in `CompareComplex`. The `CompareSequences`, `CompareUnordered` (sets / `IgnoreOrder`), `CompareDictionaries`, and multidimensional-array branches never registered or checked visited pairs, so a cycle routed through any of them never terminated.

The set / `IgnoreOrder` / dictionary-key paths recurse through `IsEquivalent`, which starts a **fresh** comparison with an empty `_visitedPairs`.

## Fix

- **Hoist the cycle/identity guard into `CompareNodes`** so every composite reference node (sequence, array, set, dictionary, complex) is tracked.
- **Add a stack-scoped `_activePairs` set** that represents the live ancestor chain, and **share it into `IsEquivalent`** so cycles that close through order-insensitive matching also terminate.

### DIFF
--- a/src/Shouldly.Tests/ShouldBeEquivalentTo/CyclicCollectionScenario.cs
+++ b/src/Shouldly.Tests/ShouldBeEquivalentTo/CyclicCollectionScenario.cs
@@ -1,0 +1,94 @@
+namespace Shouldly.Tests.ShouldBeEquivalentTo;
+
+/// <summary>
+/// Regression coverage for cycles that close through a collection, array, set or dictionary node.
+/// </summary>
+public class CyclicCollectionScenario
+{
+    [Fact]
+    public void ShouldPassWhenListContainsItself()
+    {
+        var subject = new List<object> { 1 };
+        subject.Add(subject);
+
+        var expected = new List<object> { 1 };
+        expected.Add(expected);
+
+        subject.ShouldBeEquivalentTo(expected);
+    }
+
+    [Fact]
+    public void ShouldDetectDifferencesAlongsideSelfReferentialList()
+    {
+        var subject = new List<object> { 1 };
+        subject.Add(subject);
+
+        var expected = new List<object> { 2 };
+        expected.Add(expected);
+
+        // The cycle terminates, but the differing sibling element is still reported.
+        Should.Throw<ShouldAssertException>(() => subject.ShouldBeEquivalentTo(expected));
+    }
+
+    [Fact]
+    public void ShouldPassWhenTwoListsFormMutualCycle()
+    {
+        var subjectA = new List<object>();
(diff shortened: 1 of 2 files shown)


slang25:
Great work @Bafyn 🙂  thanks for the contribution

PR #1319, opened 2026-08-17
Fix #1318: ShouldBe(ignoreOrder: true) passing for unequal collections with default value

### LINKED ISSUE #1318: `ShouldBe(expected, ignoreOrder: true)` silently passes for unequal collections when `default(T)` is present
## Summary

`ShouldBe(expected, ignoreOrder: true)` can pass even though the two collections are not equal. A failing assertion succeeds silently.

The problem occurs whenever `default(T)` is present in the expected collection and at least one actual element has no genuine match. For `int` that means a `0` in `expected`; for reference types it means a `null` in `expected`.

## Steps to reproduce

```
int[] actual = [5, 5];
int[] expected = [5, 0];
actual.ShouldBe(expected, ignoreOrder: true);   // passes
```

```
// single element mismatch, expected value is 0 == default(int)
int[] actual = [1];
int[] expected = [0];
actual.ShouldBe(expected, ignoreOrder: true);   // passes
```

```
// same problem for reference types, where default(T) is null
string?[] actual = ["a"];
string?[] expected = [null];
actual.ShouldBe(expected, ignoreOrder: true);   // passes
```

## Expected behavior

`[5, 5].ShouldBe([5, 0], ignoreOrder: true)` (and the other cases above) should throw `ShouldAssertException`, because the collections do not contain the same elements.

## Actual behavior

The assertion returns successfully and the test passes, so an unequal collection is reported as equal.

Bafyn:
Fixes #1318 

## Root cause

The order-insensitive comparison in `Is.EqualIgnoreOrder` (`src/Shouldly/Internals/Is.cs`) matches each actual element with `FirstOrDefault` and then removes the match:

```csharp
var expectedList = expected.ToList();
foreach (var actualElement in actual)
{
    var match = expectedList.FirstOrDefault(x => Equal(x, actualElement, comparer));
    if (!expectedList.Remove(match!)) // List<T>.Remove works fine when null is passed.
        return false;
}

return !expectedList.Any();
```

When no element is equivalent to `actualElement`, `FirstOrDefault` returns `default(T)`. If `default(T)` happens to be an element of `expectedList` (`0` for `int`, `null` for a reference type), `expectedList.Remove(default(T))` succeeds and wrongly consumes that element, so an unmatched actual element is treated as matched and the method can return `true` for unequal collections.

### DIFF
--- a/src/Shouldly.Tests/ShouldBe/EnumerableType/IgnoreOrderDefaultValueScenario.cs
+++ b/src/Shouldly.Tests/ShouldBe/EnumerableType/IgnoreOrderDefaultValueScenario.cs
@@ -1,0 +1,70 @@
+namespace Shouldly.Tests.ShouldBe.EnumerableType;
+
+/// <summary>
+/// Regression coverage for the order-insensitive comparison matching an unmatched element against <c>default(T)</c>.
+/// </summary>
+public class IgnoreOrderDefaultValueScenario
+{
+    [Fact]
+    public void ShouldFailWhenAnUnmatchedElementCollidesWithDefaultInt()
+    {
+        int[] actual = [5, 5];
+        int[] expected = [5, 0];
+
+        Should.Throw<ShouldAssertException>(() => actual.ShouldBe(expected, ignoreOrder: true));
+    }
+
+    [Fact]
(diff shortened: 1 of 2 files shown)


slang25:
Great find! Thanks again @Bafyn

